PR #2, opened 2018-03-20
qemuriscv64: pin kernel to 4.15

twoerner:
The 4.16 kernel doesn't currently seem to be able to come up to a login
prompt. Set the default to 4.15 (which can) for the time-being.

Signed-off-by: Trevor Woerner <twoerner@gmail.com>

### DIFF
--- a/conf/machine/qemuriscv64.conf
+++ b/conf/machine/qemuriscv64.conf
@@ -11,7 +11,7 @@ MACHINE_ARCH = "riscv64"
 KERNEL_IMAGETYPE = "vmlinux"
 
 PREFERRED_PROVIDER_virtual/kernel = "linux-riscv"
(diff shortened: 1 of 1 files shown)


kraj:
Can you try to change the baudrate and see if that fixes 4.16 builds ? something like this

```diff
-SERIAL_CONSOLES = "38400;ttyS0 38400;hvc0"
+SERIAL_CONSOLES = "115200;ttyS0 115200;hvc0"
```

mirzak:
I have experienced the same issue with the current master.

Though was working on something else and these two commits in my fork https://github.com/mirzak/meta-riscv-1/tree/clean-up-kernel somehow fix it, without knowing exactly why.

kraj:
its more likely related to baudrate, test it out and let me know

PR #7, opened 2018-04-10
Freedom u540 sd

mfischer:
Hi Khem,

first shot of pull request. Let me know if this works for you,
not sure about the separate machines + MACHINE_OVERRIDE.
Ideas?

Moritz


conf/machine/freedom-u540-common.inc
```diff
@@ -7,6 +7,7 @@ require conf/machine/include/tune-riscv.inc
 
 MACHINE_FEATURES = "screen keyboard ext2 ext3 serial"
 MACHINE_ARCH = "riscv64"
+MACHINEOVERRIDE="freedom-u540:${MACHINE}"
```
kraj: probably we should use SOC_FAMILY here

conf/machine/freedom-u540-common.inc
```diff
@@ -18,16 +19,10 @@ QEMUVERSION = "riscv"
...
+SERIAL_CONSOLE = "115200 console"
 
 MACHINE_EXTRA_RRECOMMENDS += " kernel-modules"
 
-RISCV_BBL_PAYLOAD ?= "${KERNEL_IMAGETYPE}${KERNEL_INITRAMFS}-${MACHINE}.bin"
-
-INITRAMFS_IMAGE_BUNDLE = "1"
-INITRAMFS_IMAGE = "riscv-initramfs-image"
-KERNEL_INITRAMFS = '-initramfs'
-
```
kraj: Just comment above three vars out and add a comment asking to enable them if someone wants initramfs image.

conf/machine/freedom-u540-initramfs.conf
```diff
@@ -0,0 +1,28 @@
...
+IMAGE_BOOT_FILES = ""
+
+### wic default support
+WKS_FILE_DEPENDS ?= " \
+    riscv-pk \
+    e2fsprogs-native \
+    bmap-tools-native \
+"
+
+WKS_FILE ?= "freedom-u540-bbl.wks"
```
kraj: not needed.

conf/machine/freedom-u540-common.inc
mirzak: Agreed on this. No need for two machines. Could convert above to a "config" as well, but for now I feel it is OK to comment it out with a comment what it means to un-comment it.

mfischer:
Updated PR.

conf/machine/freedom-u540.conf
```diff
@@ -18,15 +18,20 @@ QEMUVERSION = "riscv"
 
 EXTRA_IMAGEDEPENDS += "riscv-pk"
 
-SERIAL_CONSOLE = "115200 ttyS0"
+SERIAL_CONSOLE = "115200 console"
 
 MACHINE_EXTRA_RRECOMMENDS += " kernel-modules"
 
-RISCV_BBL_PAYLOAD ?= "${KERNEL_IMAGETYPE}${KERNEL_INITRAMFS}-${MACHINE}.bin"
```
kraj: you can keep this as it is. Since KERNEL_INITRAMFS is not defined it will be reduced to what you expect it to be ${KERNEL_IMAGETYPE}-${MACHINE}.bin

PR #13, opened 2018-05-15
riscv-pk: depend on virtual/kernel:do_deploy

rsalveti:
bbl_payload requires the kernel deployed artifacts (e.g. vmlinux) during
compile time in order to generate the final bbl binary, so add
virtual/kernel:do_deploy as a do_compile dependency to get the
dependency task in the expected order.

Signed-off-by: Ricardo Salveti <ricardo@opensourcefoundries.com>


rsalveti:
@kraj mind having another look?

rsalveti:
The CI build failure looks like a timeout (build taking too long), is this expected?

PR #16, opened 2018-06-22
qemu-riscv: update to the latest

rsalveti:
8dad5ee5a63889fb8cb965df23a8101b12786e68 is not available anymore in riscv-qemu, so update to the latest revision available.

Signed-off-by: Ricardo Salveti <ricardo@opensourcefoundries.com>


rsalveti:
Still testing locally, but this is the latest from riscv-qemu.

sashimi-yzh:
bb6b00ce894a445811a5bac165c9ad3e7b437379 still does not exist in riscv-qemu

alistair23:
The branch is constantly forced pushed. We should swap to just using the branch instead of hard coding a commit.

kraj:
we need to switch back to master branch.

PR #19, opened 2018-07-31
qemuriscv64: uncomment accidentally commented out QEMUVERSION

kontar:
Looks like commit 2cf991cc3191 ("add templates for commits and issues")
accidentally commented out QEMUVERSION for qemuriscv64 machine. It
broke a build because default version of QEMU gets selected that doesn't
support RISC-V.

Uncomment it back.

Signed-off-by: Taras Kondratiuk <takondra@cisco.com>

### DIFF
--- a/conf/machine/qemuriscv64.conf
+++ b/conf/machine/qemuriscv64.conf
@@ -13,7 +13,7 @@ MACHINE_FEATURES = "screen keyboard ext2 ext3 serial"
 KERNEL_IMAGETYPE = "vmlinux"
 
 GDBVERSION = "riscv"
-#QEMUVERSION = "riscv"
+QEMUVERSION = "riscv"
 
 SERIAL_CONSOLES = "115200;ttyS0 115200;hvc0"
 


alistair23:
The default version of QEMU does support RISC-V. I think moving to using mainline QEMU is a step in the right direction (although maybe a little early as QEMU 3.0 adds much better support).

What is the failure you are seeing?

kraj:
Hmm qemu from core worked ok what issues do you see in general

kontar:
gobject-introspection was failing during core-image-minimal build because it couldn't find qemu-riscv64. I've found the rootcause: poky.conf is overriding QEMU_TARGETS. I believe Poky build with default configs should be broken for you too.

```
% bitbake qemu-native -e
...
# $QEMU_TARGETS [2 operations]
#   set? /projects/risc-v/poky/meta-poky/conf/distro/poky.conf:42
#     "arm aarch64 i386 mips mipsel mips64 mips64el nios2 ppc x86_64"
#   set? /projects/risc-v/poky/meta/recipes-devtools/qemu/qemu.inc:12
#     "arm aarch64 i386 mips mipsel mips64 mips64el ppc riscv32 riscv64 sh4 x86_64"
# pre-expansion value:
#   "arm aarch64 i386 mips mipsel mips64 mips64el nios2 ppc x86_64"
QEMU_TARGETS="arm aarch64 i386 mips mipsel mips64 mips64el nios2 ppc x86_64"
```

I'm not sure why Poky needs to set QEMU_TARGETS explicitly though. For now just removed it from poky.conf. Closing this PR.

Actually if default qemu works, then it makes sense to switch all riscv machines to it. Current mix is a big confusing.

kraj:
@kontar I do not use poky distro but rather bec distro ( which i maintain ) and it inherits nodistro from oe-core so I do build riscv support automatically as it does not restrict qemu targets like poky. that said please send a patch to poky to enable it and I will vote it in.

kontar:
@kraj I'll sent a patch. I don't normally use poky either, but meta-riscv/README.md instructions pulls Poky. So if somebody follow README they should hit this issue.

kraj:
OK lets apply this patch until its fixed in poky. Can you resend

PR #25, opened 2018-08-27
Remove the RISC-V QEMU fork and update the machines

alistair23:
Remove the RISC-V QEMU fork as QEMU 3.0 has full RISC-V support.

Also build the EXT4 image as it's useful for development and update the QEMU serial arguments.


pino-kim:
Dear alistair23.
thanks to your pr work.
qemu 3.0 is full support riscv architecture, so remove qemu-riscv.bb?
Am I right?

conf/machine/qemuriscv64.conf
```diff
@@ -29,3 +29,4 @@ QB_DEFAULT_KERNEL = "bbl"
 QB_NETWORK_DEVICE = "-device virtio-net-device,netdev=net0,mac=@MAC@"
 QB_ROOTFS_OPT = "-drive file=@ROOTFS@,format=raw,id=hd0 -device virtio-blk-device,drive=hd0"
 QB_SLIRP_OPT = "-netdev user,id=net0,hostfwd=tcp::22222-:22"
+QB_OPT_APPEND = "-nographic -serial mon:stdio -serial null"
```
kraj: Can you explain the reasoning behind this change a bit ?

alistair23: The default args for QEMU will create a display so by adding -nographic we will QEMU to run in text only mode and then after that we just tell QEMU to display the serial output and the QEMU monitor the the users stdio.

It's just being more explicit with the console output.

kraj: we do have options to control it from runqemu tool
```
runqemu nographic serial 
```
This change sort of overrides the tool behavior, would this
runqemu cmd suffice for your needs ?

alistair23: That would work, but I think that should be the default. Is it possible to make that the default?

kraj: all qemu targets in OE-Core can boot into graphics hence the default. If we can convince the oe-core community I would suggest to make this change there. I personally have no hard opinion about the default. but I want RISC-V to follow the
oe-core

alistair23: Ok, I've dropped this patch from the PR.

PR #27, opened 2018-08-28
Update to the latest RISC-V fork of GDB and add the bison dependency

alistair23:


recipes-devtools/gdb/gdb_riscv.bb
```diff
@@ -1,6 +1,8 @@
 require recipes-devtools/gdb/gdb.inc
 require gdb-${PV}.inc
 
+DEPENDS += "bison-native"
```
kraj: if we need bison-native for target recipe here, then we should need it in OE-Core as well. Please send a different pull for fixing it in OE-Core, until then we can apply this here.

PR #43, opened 2018-10-08
u-boot: Use mainline u-boot

alistair23:
Now that QEMU RISC-V virt machine support is in master u-boot use the
mainline u-boot instead of a fork.


recipes-bsp/u-boot/u-boot_%.bbappend
```diff
@@ -1,5 +1,6 @@
-SRC_URI_riscv64 = "git://git.denx.de/u-boot-x86.git;branch=riscv-working;rebaseable=1"
+# Remove patches applied in the common .bb
+SRC_URI_riscv64 = "git://git.denx.de/u-boot.git"
```
kraj: if its upstream then do we need to override SRC_URI ? I guess just SRCREV setting should be enough
then we can also get the upstream OE fixes along.

alistair23: We need to not apply the additional patches as they have already been applied in the master branch

PR #44, opened 2018-10-21
qemuriscv64: Express u-boot bootloader choice

kraj:
In a multi-BSP env u-boot may be provided by other BSP layers as well so
we need to be explicit about u-boot we need

### DIFF
--- a/conf/machine/qemuriscv64.conf
+++ b/conf/machine/qemuriscv64.conf
@@ -5,6 +5,9 @@
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-riscv"
 PREFERRED_VERSION_linux-riscv ?= "4.18%"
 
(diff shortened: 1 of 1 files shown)


conf/machine/qemuriscv64.conf
```diff
@@ -5,6 +5,9 @@
 PREFERRED_PROVIDER_virtual/kernel ?= "linux-riscv"
 PREFERRED_VERSION_linux-riscv ?= "4.18%"
 
+PREFERRED_PROVIDER_virtual/bootloader ?= "u-boot"
+PREFERRED_PROVIDER_u-boot ?= "linux-riscv"
```
pino-kim: hello @kraj
I think PREFERRED_PROVIDER_u-boot parament value is "u-boot" (line:9)
Can you check it again please.

PR #45, opened 2018-11-05
riscv-pk: Update to the latest version

alistair23:
Update to the latest version of the RISC-V Proxy Kernel. This means we
are running an updated bbl.


recipes-devtools/riscv-tools/files/0002-Makefile.in-Don-t-specify-a-march-if-one-isn-t-speci.patch
```diff
@@ -0,0 +1,39 @@
...
+-# compile it with a no-float ABI. i.e., ilp32 for 32-bit and lp64 for 64-bit.
+-
+-ifeq (@WITH_ARCH@,)
+-  march := -march=$(shell @CC@ -v 2>&1 | \
+-                    sed 's/ /\n/g' | grep with-arch | awk -F= '{print $$2}')
+-else
+-  march := -march=@WITH_ARCH@
+-endif
+-is_32bit := $(findstring 32,$(march))
+-mabi := -mabi=$(if $(is_32bit),ilp32,lp64)
```
kraj: is there a way to make it upstreamable way ?

alistair23: Not that I can think of. We have to ensure it's not set and in the FreedomSDK environment they need it to be set.

PR #49, opened 2018-11-26
gdb: Update to mainline GDB

alistair23:
The mainline master GDB is very good now and even includes support for
debugging RISC-V Linux applications. Use the master branch until the
next GDB release is cut with all of this new support.


recipes-devtools/gdb/gdb_riscv.bb
```diff
@@ -3,8 +3,6 @@ require gdb-${PV}.inc
 
 DEPENDS += "bison-native"
 
-SRC_URI += "file://0001-linux-ptrace-Fix-RISC-V-cross-compilation-error.patch"
```
kraj: Please git rm ```0001-linux-ptrace-Fix-RISC-V-cross-compilation-error.patch``` as well

alistair23: Ah, that was stupid. Fixed.

PR #52, opened 2018-11-27
libffi: Add libffi 3.3-rc0

alistair23:
libffi 3.3-rc0 has much better RISC-V suppor then our single patch. Use
the RC release instead of a single patch.


alistair23:
Fixed!

PR #55, opened 2018-11-29
qemuriscv64: Remove GDBVERSION property

alistair23:


conf/machine/qemuriscv64.conf
```diff
@@ -15,8 +15,6 @@ MACHINE_FEATURES = "screen keyboard ext2 ext3 serial"
 
 KERNEL_IMAGETYPE = "vmlinux"
 
-GDBVERSION = "riscv"
```
kraj: this would make it use gdb from oe-core, we dont want that yet

alistair23: Argh, you are right. I jumped the gun here.

PR #56, opened 2018-11-29
Add support for the qemuriscv32 machine

alistair23:
Add support for the 32-bit QEMU machine.


recipes-core/glibc/glibc-initial_%.bbappend
```diff
@@ -0,0 +1,18 @@
...
+# Force it to use the RISC-V fork
+
+GLIBC_GIT_URI_qemuriscv32 = "git://github.com/riscv/riscv-glibc.git"
+SRCBRANCH_qemuriscv32 = "32bit"
+SRCREV_qemuriscv32 = "83d5eaa75361ba0cb7969669452c9ff233ac710b"
+
+SRC_URI_remove_qemuriscv32 = " \
+    file://0031-sysdeps-ieee754-prevent-maybe-uninitialized-errors-w.patch \
+    file://0032-sysdeps-ieee754-soft-fp-ignore-maybe-uninitialized-w.patch \
+    file://0033-locale-prevent-maybe-uninitialized-errors-with-Os-BZ.patch \
```
kraj: I dont see these patches added, where will they come from ?

conf/machine/qemuriscv32.conf
```diff
@@ -0,0 +1,61 @@
+#@TYPE: Machine
```
kraj: can we create a ```riscv-common.inc``` and move common bits into it from 32bit and 64bit conf files.

recipes-core/glibc/glibc-initial_%.bbappend
alistair23: These are being removed, or at least not applied as they don't apply to the newer tree needed for 32bit support

conf/machine/qemuriscv32.conf
alistair23: Yep

alistair23: Updated in the new version

PR #59, opened 2018-12-03
libgpg-error: Add support for RISC-V 32-bit

alistair23:
This adds support for 32-bit builds to libgpg-error. Upstream status is pending company approval.

Thanks for @pino-kim for helping figure out the patch.

### DIFF
--- a/recipes-support/libgpg-error/libgpg-error_%.bbappend
+++ b/recipes-support/libgpg-error/libgpg-error_%.bbappend
@@ -1,0 +1,5 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+SRC_URI_append_riscv32 = "\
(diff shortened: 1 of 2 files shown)


recipes-support/libgpg-error/libgpg-error_%.bbappend
```diff
@@ -0,0 +1,5 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
+
+SRC_URI_append_qemuriscv32 = "\
```
kraj: this should be ```riscv32``` so it can apply to all machines and not just to qemu

alistair23: Fixed in updated patch

PR #61, opened 2018-12-07
linux: Add the 4.20-rc5 RISC-V kernel

alistair23:
In preperation for the 4.20 kernel release (and to help test) add the
4.20 kernel but don't set it as the default yet.


recipes-kernel/linux/linux-riscv_4.20.bb
```diff
@@ -0,0 +1,8 @@
+require recipes-kernel/linux/linux-riscv-common.inc
```
kraj: Can you also add `DEFAULT_PREFERENCE = "-1" `  for this please ?

alistair23: Fixed

PR #70, opened 2018-12-14
machine: Allow overriding TCLIBC

alistair23:

### DIFF
--- a/conf/machine/freedom-u540.conf
+++ b/conf/machine/freedom-u540.conf
@@ -38,8 +38,6 @@ WIC_CREATE_EXTRA_ARGS ?= "--no-fstab-update"
 
 IMAGE_BOOT_FILES = ""
 
-TCLIBC = "glibc"
-
 ### wic default support
 WKS_FILE_DEPENDS ?= " \
     riscv-pk \
--- a/conf/machine/include/qemuriscv.inc
+++ b/conf/machine/include/qemuriscv.inc
@@ -14,7 +14,6 @@ GDBVERSION = "riscv"
 
 SERIAL_CONSOLES = "115200;ttyS0"
 
-TCLIBC = "glibc"
 IMAGE_FSTYPES_append = " ext4"
 
 MACHINE_EXTRA_RRECOMMENDS += " kernel-modules"


conf/machine/freedom-u540.conf
```diff
@@ -38,7 +38,7 @@ WIC_CREATE_EXTRA_ARGS ?= "--no-fstab-update"
 
 IMAGE_BOOT_FILES = ""
 
-TCLIBC = "glibc"
+TCLIBC ?= "glibc"
```
kraj: musl does not work for riscv64 yet. Thats why it is locked to glibc. What is the usecase here ?

conf/machine/baremetal-riscv32.conf
```diff
@@ -23,7 +23,7 @@ USER_CLASSES_remove = "image-prelink"
 #TARGET_LINK_HASH_STYLE = ""
 LINKER_HASH_STYLE = ""
 
-TCLIBC = "baremetal"
+TCLIBC ?= "baremetal"
```
kraj: we may want to define another machine config, baremetal is for generating pure baremetal toolchain

conf/machine/include/qemuriscv.inc
```diff
@@ -14,7 +14,7 @@ GDBVERSION = "riscv"
 
 SERIAL_CONSOLES = "115200;ttyS0"
 
-TCLIBC = "glibc"
+TCLIBC ?= "glibc"
```
kraj: we do not support musl yet, thats the reason to hardcode

conf/machine/freedom-u540.conf
alistair23: Not allowing an override is confusing as if people set it they expect the override to work not just be ignored. If that causes a failure then so be it, if it silently works that is just confusing. There is a [musl port](https://github.com/riscv/riscv-musl) so maybe people want to use it.

conf/machine/baremetal-riscv32.conf
alistair23: Same point as above, it's confusing when the override doesn't work

conf/machine/include/qemuriscv.inc
alistair23: Same point as the freedome-u540 change

conf/machine/freedom-u540.conf
kraj: Yes, thats fine.putting TCLIBC here was done for ease of use and automation. I can change my workflow. Please remove it completely from here, since its a DISTRO property anyway.

conf/machine/baremetal-riscv32.conf
kraj: baremetal is baremetal. I dont think musl/glibc argument applies here.  Probably its better to post a new .conf file which takes care of other TCLIBC options. e.g. newlib

conf/machine/include/qemuriscv.inc
kraj: Remove it completely.

conf/machine/freedom-u540.conf
alistair23: Removed

conf/machine/baremetal-riscv32.conf
alistair23: Ok, I'll leave this as is for now.

PR #71, opened 2018-12-19
glibc-initial_%.bbappend: Drop

kraj:
glibc-initial has been removed from oe-core in toolchain build step


kraj:
Thanks for testing.
yes this is expected, the corresponding OE-Core changes are still in master-next, thats why this pull is still parked.

PR #80, opened 2019-01-03
Use the linux-yocto kernel for the 64-bit QEMU RISC-V machine

alistair23:
Try to use the mainline kernel where possible.


alistair23:
This doesn't build, we need to specify RISC-V as a supported COMPATIBLE_MACHINE and we also need to deal with the defconfig for qemuriscv64/32

kraj:
I will wait for v2

alistair23:
This started as just a WIP for comments to see if other agreed this is the right direction, but now I think it's ok to be a patch set.

It changes the qemuriscv64 machine to use the linux-yocto kernel. Hopefully we can also convert qemusicv32 for the 4.20 kernel as the 4.19 linux-yocto kernel doesn't seem to work.

kraj:
I dont know if it is right approach for every machine, if we have to move qemuriscv64 to oe-core then yocto kernel maintainer has to agree to maintain riscv port like other qemu machines for yocto. Since linux-yocto != upstream kernel, it certainly would make sense to have such an option but I would think it still is better to keep pure upstream recipe in this layer. People may not want to use linux-yocto

kraj:
Yes doing that for qemu machines is ok.

alistair23:
I'm going to close this and I'll just try to have qemuriscv64 support merged into the linux-yocto kernel.

kraj:
I think that would be great, once we have that then we can keep this kernel just for real machines.

PR #82, opened 2019-01-10
recipes-kernel: Update the Freedom U540 Linux defconfig

alistair23:
Update the Freedom U540 defconfig to work with:
 - AMD GPUs attached with PCIe MicroSemi Expansion board
    - Including Wayland and X11 display via HDMI
 - USB input attached with PCIe MicroSemi Expansion board

At the same time we are also swapping to use the config output of
savedefconfig instead of copying the entire .config file. This
simplifies the defconfig management.


kraj:
while no issues with this pull, I see that we have stored a generated defconfig into metadata, I think what we should have is the output of `make savedefconfig` which is actually the source  while you are at it and can validate it as well. Would you mind generating the correct defconfig and post it as a patch along with this change in the pull ?

alistair23:
I'm not sure what you mean? This patch updated the defconfig file stored in meta-riscv to be the output of make savedefconfig

alistair23:
This patch is changing the defconfig to be the output of savedefconfig and not use the .config. I'm don't understand what else needs to be done.

kraj:
if it is output of `make savedefconfig` we might be fine. Although it does not look like one. Usually defconfigs are pretty small when generated using savedefconfig

alistair23:
The reason it's so long is because we enable a large number of USB, HID, Sound Card and input drivers.

I enabled these in getting all display, keyboard, mouse and audio working on the board. I'm sure not all of these are required, but I have left them enabled to ensure that others can have their devices work.

Maybe we can look at removing some of them if required.

alistair23:
https://github.com/riscv/meta-riscv/pull/83 removes some of the configs.

PR #83, opened 2019-01-11
recipes-kernel: Remove extra components from the Freedom U540 config

alistair23:


kraj:
thanks for cleanup

PR #84, opened 2019-01-14
qemuriscv64: Don't use the linux-riscv kernel

alistair23:
Default to the mainline 4.19 kernel for the 64-bit QEMU machine.


alistair23:
I don't think we need any other changes

PR #85, opened 2019-01-17
layer.conf: Do not lock to gcc

kraj:
clang 8.0 does have risc-v backend


alistair23:
Have you tested the RISC-V support in clang?

kraj:
it will be first supported in upcoming clang8, so this is groundwork for that, I have clang8 recipes already available on kraj/clang-8 branch in meta-clang

alistair23:
Cool!

PR #87, opened 2019-01-20
qemuriscv64.conf: Pin to linux-riscv again

kraj:
linux-yocto hangs before it reaches console, until that is
fixed we can not switch

### DIFF
--- a/conf/machine/qemuriscv64.conf
+++ b/conf/machine/qemuriscv64.conf
@@ -7,3 +7,6 @@ require conf/machine/include/qemuriscv.inc
 EXTRA_IMAGEDEPENDS += "u-boot"
 UBOOT_MACHINE = "qemu-riscv64_defconfig"
 UBOOT_ELF = "u-boot"
+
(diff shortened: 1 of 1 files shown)


alistair23:
I noticed that the console isn't displayed in Linux-yocto-dev.

I do have some patches to improve the console, I'll look into this as well.

alistair23:
I can't reproduce this. The console isn't printed after earlyboot (sending a patch today) but I don't see any hangs. Can you provide more details?

kraj:
Yeah it doesn’t look like a hang per se but it’s console not coming up and freeze

alistair23:
Ah ok, that's just a defconfig issue. I have sent patches to Bruce

kraj:
Alright let me know when that’s merged in master and we can switch back the kernels after validation

alistair23:
Bruce sent the patches out today. I have a big kernel series I'll push out tonight, we can merge it when Bruce's patch propagates into meta.

kraj:
sounds good. Send me a ping when OE-Core lands it

PR #89, opened 2019-01-24
Kernel improvements

alistair23:
This patch set upgrades the kernels available.


alistair23:
This is ready to go now

alistair23:
It looks like the config changes break the 4.19 build. Sorry about that, I'll send a patch soon

PR #93, opened 2019-01-31
openSBI: Initial commit

alistair23:
This adds openSBI support. openSBI aims to be the reference
implementation for the M mode firmware. We are going to use it to remove
BBL from the boot flow.

The eventual aim is to have openSBI merged into the meta layer to
support qemuriscv* machines.

Unfortunately openSBI requires a 5.0 kernel, so at the moment we aren't
going to use it.


alistair23:
Can we merge this one? Even without using it by default it would be nice to have as an option

palmer-dabbelt:
I'm very happy to get rid of BBL, but I think it's Khem's call.

alistair23:
This PR just adds openSBI. It doesn't use it or get rid of BBL. Once the 5.0 kernel is out and 32bit support is added to openSBI we can swap across to openSBI by default. After that we can look at removing BBL.

kraj:
Makes sense, I think we can keep it disabled for now.

PR #100, opened 2019-02-22
glibc: Remove a patch that doesn't apply to the 32-bit fork

alistair23:


alistair23:
Patch "glibc: Remove ChangeLog diff from patch" sent to the mailing list

PR #104, opened 2019-02-26
opensbi: Bump to version 0.3

alistair23:


recipes-bsp/opensbi/opensbi_0.3.bb
```diff
@@ -10,11 +10,10 @@ inherit autotools-brokensep
 
 SRC_URI = "https://github.com/riscv/opensbi/archive/v${PV}.tar.gz \
```
kraj: Usually I would prefer a proper release or using git URI, github rebuilds the archives and it can end up in updated checksums silently.

alistair23: I can change it to a git clone if that's better. At the moment the "official" release is via the GitHub archives though

kraj: I think its better to use SHA1 and  git fetcher instead

alistair23:
Updated.

PR #106, opened 2019-03-04
freedom-u540: Update to 5.0 kernel and OpenSBI

alistair23:


kraj:
does 5.0 work on qemu ?

alistair23:
The 5.0 kernel does work on QEMU. For the QEMU machines I want to use linux-yocto 5.0, which is why they aren't being upgraded at the moment.

pino-kim:
@ali

pino-kim:
@alistair23 @kraj 
I wonder kernel we use in recipes. 
Is it defer kernel.org's kernel and yocto's kernel?(Not a update version.)

alistair23:
My eventual goal is to use linux-yocto for all machines. This is regularly updated with bug and security fixes. I think a well maintained version of the kernel will be critical for anyone deploying a product.

In the mean time the HiFive Unleashed needs a fork of the kernel to address missing upstream work

kraj:
@alistair23 i think upstream kernel.org stable releases is better option they are maintained by Greg, linux-yocto does have its own set of patches which are not upstream and workflow is not quite common,
so I think we should concentrate on k.o more than anything

pino-kim:
@kraj @alistair23 
Hmm ... I think both you right. 
I suggest make kernel bbfile one is for yocto-linux and one is for k.o .
Someone want to use yocto-linux or k.o I think. 

By the way can I use current stable 5.0 with qemu and opensbi?

alistair23:
Ok, so do you think we should add a linux-mainline recipe and use that? That can just auto rev from the stable branch I guess, to avoid a maintenance burden.

With the 5.1 release we can probably do that and just add the Unleashed patches manually. For 5.0 we have too many patches right now.

The linux-riscv branch is more or less mainline now, maybe it's worth renaming?

alistair23:
Yep, 5.0 works with QEMU and OpenSBI. The QEMU machines target linux-yocto and I'm hoping to get them into the meta layer once linux-yocto 5.0 is out

pino-kim:
I think it will be soon. Next linux yocto default is 5.0 and it's work will happen near week. Bruce comment.

PR #109, opened 2019-03-07
qemuriscv: Add OpenSBI support and use the linux-yocto 5.0 kernel

alistair23:


kraj:
This looks good, have you tried running -ctestimage for qemuriscv64 ? somehow it does not detect the backdoor serial the ptest opens and tried to talk to qemu over that.

to try it use

```
TESTIMAGE_AUTO_qemuall = "1"
TEST_TARGET_IP_qemuall = "192.168.7.2"
TEST_QEMUBOOT_TIMEOUT = "50"
```

then run
```
bitbake core-image-base
```

you will see that when it tries to run tests it bails out with messages saying it cant infer if system
booted ( reached cmd prompt ) or not.

alistair23:
I haven't tried that. Is this something that used to work and now broke or has it never worked?

kraj:
no it never worked thus far and I was hoping that it will work with linux-yocto kernel and SBI switch
this would be needed to fix if we have to make RISCV a first class citizen in OE-Core

alistair23:
Ok, I can look into it. Probably not this week though as I'm at Scale

PR #110, opened 2019-03-08
Fix LTP and opensbi builds

kraj:


recipes-extended/ltp/ltp_%.bbappend
```diff
@@ -0,0 +1,4 @@
+FILESEXTRAPATHS_prepend := "${THISDIR}/${PN}:"
+
+SRC_URI_append_riscv64 = " file://0001-ptrace-Do-not-include-asm-ptrace.h-on-riscv.patch" 
+SRC_URI_append_riscv = " file://0001-ptrace-Do-not-include-asm-ptrace.h-on-riscv.patch"
```
alistair23: Don't you mean riscv32 here?

kraj: yes but I never tested on rv32, maybe I should just delete it.

PR #120, opened 2019-03-18
glibc: Update 32-bit glibc

alistair23:
We can also remove the 32-bit fix that was included in the kernel.


alistair23:
This is great! It looks like with the latest glibc submission we don't need any reverts applied to the kernel. So now QEMU 32-bit and 64-bit can use mainline. Tested in my set up

PR #122, opened 2019-03-20
recipes-kernel: linux-mainline: Initial commit of mainline 5.1 kernel

alistair23:
Add a linux-mainline kernel from 5.1-rc1. The idea is that this kernel
will autorev the 5.1 development cycle and then be updated to autorev
the 5.1 stable branch after release.


alistair23:
This is on top of https://github.com/riscv/meta-riscv/pull/120

PR #124, opened 2019-03-25
qemuriscv.inc: use ttyS0 as serial console

chinhuat:
Fix /etc/inittab issue seen when running "runqemu nographic serial":

  INIT: Id "AMA0" respawning too fast: disabled for 5 minutes

Signed-off-by: Chin Huat Ang <chinhuat@gmail.com>


alistair23:
Although this looks ok, I just realised that we are printing the entire kernel log twice. We should probably fix that at the same time.

kraj:
@alistair23 those are two different issues, lets try to serialize them.

alistair23:
Yep, was just about to comment. This looks good. It looks like the "earlycon=sbi" is causing the multiple kernel prints. Now we have to decide if we remove it and/or fix it in the kernel/OpenSBI. I'll keep digging.

PR #135, opened 2019-04-18
Add support for u-boot auto TFTP boot

alistair23:
This patch series does a few things:

1. Updates the sdcard image to include the standard /boot/ directory. This is common for OE built SD card images. At the moment though as u-boot doesn't support SD/MMC access we can't load the images from here.
2. Add support for SMP boot on the FreedomU540
3. Build u-boot by default and make it easy to set an override to include it as an OpenSBI payload.
4. Setup scripts to help with a TFTP autoboot.

### DIFF
--- a/recipes-bsp/u-boot/u-boot_2019.04.bbappend
+++ b/recipes-bsp/u-boot/u-boot_2019.04.bbappend
@@ -18,4 +18,29 @@ SRC_URI += "file://0001-riscv-add-infrastructure-for-calling-functions-on-ot.pat
             file://0016-riscv-dts-ae350-support-SMP.patch \
             file://0017-riscv-ae350-enable-SMP.patch \
             file://0018-riscv-dts-fix-CONFIG_DEFAULT_DEVICE_TREE-failure.patch \
+            file://0001-sifive-fu540-Set-default-arguments-to-help-booting.patch \
+            file://tftp-boot.txt \
            "
+
+DEPENDS += "u-boot-tools-native"
+
+UBOOT_ENTRYPOINT = "0x80200000"
+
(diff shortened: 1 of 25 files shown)


recipes-bsp/u-boot/u-boot_2019.04.bbappend
```diff
@@ -18,4 +18,36 @@ SRC_URI += "file://0001-riscv-add-infrastructure-for-calling-functions-on-ot.pat
...
+
+DEPENDS += "u-boot-tools-native"
+
+UBOOT_ENTRYPOINT = "0x80200000"
+
+# Overwrite this for your server
+TFTP_SERVER_IP ?= "127.0.0.1"
+
+u_boot_do_patch() {
+      sed -i -e 's,@SERVERIP@,${TFTP_SERVER_IP},g' ${WORKDIR}/0001-sifive-fu540-Set-default-arguments-to-help-booting.patch
```
kraj: Can we do it after patching ? say via do_configure_prepend, editing patches seems quite cryptic to me.

alistair23: I did think of that, but that seems just as messy. If you prefer that I'll give it a shot.

kraj: hmm how many such instances are there ?

alistair23: Just 1, it actually seems easy. Just testing it now

alistair23:
V2 sent

alistair23:
Is this ok to be merged?

PR #145, opened 2019-06-20
Update with qemuriscv64 in OE-Core

alistair23:
Now that the qemuriscv64 machine is in OE-Core let's update the
meta-riscv layer to use that.


alistair23:
Ping?

PR #147, opened 2019-06-27
Linux 5.2-rcX from master

tmagik:
  
Use the 5.2-rcX series kernels with new (stable) device tree format(s)

issues in the first part.**
  
**-Tag line(s) at the end.**

Signed-off-by: Troy Benjegerdes <troy.benjegerdes@sifive.com>

### DIFF
--- a/recipes-kernel/linux/linux-mainline_5.2.bb
+++ b/recipes-kernel/linux/linux-mainline_5.2.bb
@@ -1,9 +1,9 @@
 require recipes-kernel/linux/linux-mainline-common.inc
 
-LINUX_VERSION ?= "5.1.x"
+LINUX_VERSION ?= "5.2.x"
 KERNEL_VERSION_SANITY_SKIP="1"
 
-BRANCH = "linux-5.1.y"
+BRANCH = "master"
 SRCREV = "${AUTOREV}"
 SRC_URI = " \
     git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git;branch=${BRANCH} \
@@ -11,24 +11,17 @@ SRC_URI = " \
 
 # Out of tree patches required for HiFive Unleashed
 SRC_URI_append_freedom-u540 = " \
-    file://0001-tty-serial-add-driver-for-the-SiFive-UART.patch \
-    file://0002-gemgxl-mgmt-implement-clock-switch-for-GEM-tx_clk.patch \
     file://0003-u54-prci-driver-for-core-U54-clocks.patch \
     file://0004-u54-prci-driver-for-core-U54-clocks.patch \
     file://0005-gpio-sifive-support-GPIO-on-SiFive-SoCs.patch \
     file://0006-pwm-sifive-add-a-driver-for-SiFive-SoC-PWM.patch \
-    file://0007-RISC-V-Networking-fix-Hack.patch \
     file://0008-pcie-microsemi-Add-support-for-the-Vera-board-root-c.patch \
     file://0009-HACK-Revert-of-device-Really-only-set-bus-DMA-mask-w.patch \
     file://0010-HACK-radeon-Don-t-set-PCI-DMA-mask.patch \
 "
 
-# Fix a breakage with the current 32bit glibc fork
-SRC_URI_append_qemuriscv32 = " file://0001-Revert-riscv-Use-latest-system-call-ABI.patch"
-
-# qemu uses in-tree defconfig
-# freedom-u540 uses out-of-tree defconfig
-SRC_URI_append_freedom-u540 = " file://defconfig"
+# use in-tree defconfig
 
 KBUILD_DEFCONFIG_qemuriscv32 = "rv32_defconfig"
 KBUILD_DEFCONFIG_qemuriscv64 = "defconfig"
+KBUILD_DEFCONFIG_freedom-u540 = "defconfig"


alistair23:
Instead of replacing the stable 5.1 branch can you create a new recipe for the 5.2 kernel? While it is in RC it should not be the default.

alistair23:
Instead of replacing the stable 5.1 branch can you create a new recipe for the 5.2 kernel? While it is in RC it should not be the default.

recipes-kernel/linux/linux-mainline_5.2.bb
```diff
@@ -1,34 +1,27 @@
 require recipes-kernel/linux/linux-mainline-common.inc
 
-LINUX_VERSION ?= "5.1.x"
+LINUX_VERSION ?= "5.2.x"
 KERNEL_VERSION_SANITY_SKIP="1"
 
-BRANCH = "linux-5.1.y"
+BRANCH = "master"
```
alistair23: We should not be autoreving from master, it must be the stable branch or a SHA

recipes-kernel/linux/linux-mainline_5.2.bb
```diff
@@ -1,34 +1,27 @@
...
     file://0005-gpio-sifive-support-GPIO-on-SiFive-SoCs.patch \
     file://0006-pwm-sifive-add-a-driver-for-SiFive-SoC-PWM.patch \
-    file://0007-RISC-V-Networking-fix-Hack.patch \
     file://0008-pcie-microsemi-Add-support-for-the-Vera-board-root-c.patch \
     file://0009-HACK-Revert-of-device-Really-only-set-bus-DMA-mask-w.patch \
     file://0010-HACK-radeon-Don-t-set-PCI-DMA-mask.patch \
 "
 
-# Fix a breakage with the current 32bit glibc fork
-SRC_URI_append_qemuriscv32 = " file://0001-Revert-riscv-Use-latest-system-call-ABI.patch"
```
alistair23: Are you sure this isn't required? I don't think the 64-bit time_t has been reverted

recipes-kernel/linux/linux-mainline_5.2.bb
```diff
@@ -1,34 +1,27 @@
...
-BRANCH = "linux-5.1.y"
+BRANCH = "master"
 SRCREV = "${AUTOREV}"
 SRC_URI = " \
     git://git.kernel.org/pub/scm/linux/kernel/git/stable/linux.git;branch=${BRANCH} \
 "
 
 # Out of tree patches required for HiFive Unleashed
 SRC_URI_append_freedom-u540 = " \
-    file://0001-tty-serial-add-driver-for-the-SiFive-UART.patch \
```
alistair23: If this is removed don't we need the upstream drivers? That means we need the upstream device tree, how is that being passed to the kernel?

alistair23: > ****
> Use the 5.2-rcX series kernels with new (stable) device tree format(s)

Won't this break u-boot as it has the new bindings? How did you test this?

alistair23:
> Use the 5.2-rcX series kernels with new (stable) device tree format(s)

Won't this break u-boot as it has the new bindings? How did you test this?

davidlt:
I think, this should come with OpenSBI and U-Boot changes.
For freedom-u540 machine DTB from active kernel (virtual/kernel) needs to be embedded into OpenSBI IIRC. How do we handle that?

If we use "master" branch maybe this should be called `linux-master_git.bb`? Rebuilding this package would result in a new revision being pulled in.

For the 5.2 I would directly use one of the latest tags instead.

alistair23:
So my plan was to update OpenSBI, u-boot and the kernel. I haven't tried but I was under the impression that Linux 5.2 required at least u-boot 2019.07 and probably OpenSBI 0.4. Then we just point the OpenSBI build process to use the kernel DTB, which it already somewhat does for the expansion board. If all those updates aren't required and we can just update the kernel then we can just do that instead.

> If we use "master" branch maybe this should be called `linux-master_git.bb`? Rebuilding this package would result in a new revision being pulled in.

Use dev name, like OE-Core does. linux-mainline-dev.bb.

> For the 5.2 I would directly use one of the latest tags instead.

Or the stable branch is fine, that way we don't need to manually update all the 5.2.x releases for small fixes and security updates.

davidlt:
Fedora/RISCV uses 5.2-rc5+ + OpenSBI master + U-Boot latest RC and that works for me. I just got a report that some (randomly) cores do not come online in QEMU setup. I still haven't verified that on my setup (but will do that later today):

 [    1.426743] CPU1: failed to come online
 [    2.752426] CPU6: failed to come online

alistair23:
Yep, so that will be Linux 5.2, OpenSBI 0.4 and u-boot 2019.07. That is what I was planning on using.

pino-kim:
I run test PR test.
Qemuriscv64 boots up fine.
But Qemuriscv32 hang on a kernel panic.
leave a dmesg here

Assertion 'clock_gettime(map_clock_id(clock_id), &ts) == 0' failed at ../git/src/basic/time-util.c:55, function now(). Aborting.
[    0.983733] Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000005
[    0.984196] CPU: 0 PID: 1 Comm: init Not tainted 5.2.0-rc6 #1
[    0.984530] Call Trace:
[    0.984817] [<c003091e>] walk_stackframe+0x0/0xa6
[    0.985085] [<c0030a80>] show_stack+0x28/0x32
[    0.985310] [<c05b0352>] dump_stack+0x62/0x7e
[    0.985521] [<c003519e>] panic+0xdc/0x23e
[    0.985722] [<c0036e3c>] do_exit+0x71c/0x780
[    0.985945] [<c00378f8>] do_group_exit+0x2a/0xb8
[    0.986158] [<c003f78c>] get_signal+0x11c/0x6c8
[    0.986374] [<c0030222>] do_notify_resume+0x58/0x2e8
[    0.986607] [<c002f8ec>] ret_from_exception+0x0/0xc
[    0.987233] ---[ end Kernel panic - not syncing: Attempted to kill init! exitcode=0x00000005 ]---

----------------------
[qemuriscv32_kv5_2_rc6_dmesg.log](https://github.com/riscv/meta-riscv/files/3340216/qemuriscv32_kv5_2_rc6_dmesg.log)

alistair23:
> Qemuriscv64 boots up fine.

Cool!

> But Qemuriscv32 hang on a kernel panic.

This will need to be fixed.

Did you test on the Unleashed? It's by default the only machine using the forked kernel and the one where all the issues will arise with the recent changes

pino-kim:
@alistair23 
>Did you test on the Unleashed? It's by default the only machine using the forked kernel and the one Where all the issues will arise with the recent changes

Poorly I don't have Unleashed.  
So first check both 2 QEMU Machine boot up. 
I hope checks this PR for Unleashed.

alistair23:
No worries, thanks for testing the QEMU machines anyway. At least we know now that not everything works.

alistair23:
OpenSBI v0.4 has been released and U-Boot 2019.07 comes out in five days. Once they have been updated we can look at upgrading the kernel.

alistair23:
I have a [meta-riscv branch](https://github.com/alistair23/meta-riscv/tree/alistair/linux-5.2) that works with the 5.2 kernel. It relies on a OpenSBI and U-Boot update which I'll send next week once the U-Boot and kernel releases are out

kraj:
@alistair23 will you take care of squashing this pull along with other updates that you plan above ?

alistair23:
Yep, I have all the patches ready. Just waiting on the u-boot release.

alistair23:
I have sent all of the OE-Core patches and pushed a PR and two draft PRs to meta-riscv. This PR can be closed now.

kraj:
its addressed with https://github.com/riscv/meta-riscv/pull/153

PR #149, opened 2019-07-03
travis: Try to stop builds timing out

alistair23:
Instead of having a Travis CI system that constantly timesout let's do
two things:
 1. Adjust the Travis CI setup to be as fast as possible. By removing as
 many commands as we can, avoiding installing too many dependencies and
 doing a shallow git clone we can save time.
 2. Use timeout to test each major RISC-V machine for 55 minutes. If we
 run until we hit the timeout we count the test as a pass. This should
 allow use to rely on the tests passing before merging commits while
 also testing all three boards.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -4,24 +4,26 @@ language: shell
 compiler:
 env:
 
+git:
+  depth: 3
+
 before_install:
 
 install:
-- sudo apt-get update -qq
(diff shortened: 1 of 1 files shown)


pino-kim:
So every PR upcoming, Travis can check it. It's nice

PR #153, opened 2019-07-08
Upgrade Linux to 5.2

alistair23:
Update to the 5.2 kernel release. This includes a large number of upstream patches for the Unleashed. Due to the large defconfig changes let's drop the 5.1 support when we upgrade.


alistair23:
This needs OpenSB 0.4 and U-Boot 2019.07 merged into OE-Core before it can be merged.

davidlt:
Could you add `CONFIG_BLK_DEV_NBD=y` to allow booting using NBD (alternative to NFS) over network?

alistair23:
@davidlt I have added that.

PS, I was thinking of swapping across to the in tree defconfig for 5.3, maybe add extra configs as a fragment if required.

alistair23:
This is ready to be merged.

alistair23:
I have updated the Linux recipe to track the stable branch

Issue #156, opened 2019-07-11
Sanity check fails for qemuriscv32 machine


palmer-dabbelt:
**Description**

I'm trying to build a 32-bit setup and I'm getting an unexpected error message.  I've had this work before but I can't figure out what I'm doing differently this time around.

**Steps to reproduce the issue:**
1. Create a new repo according to the README.
2. Modify auto.conf to target qemuriscv32 instead of qemuriscv64
3. `bitbake core-image-minimal`

**Describe the results you received:**

````
$ bitbake core-image-minimal
ERROR:  OE-core's config sanity checker detected a potential misconfiguration.
    Either fix the cause of this error or at your own risk disable the checker (see sanity.conf).
    Following is the list of potential problems / advisories:

    MACHINE=qemuriscv32 is invalid. Please set a valid MACHINE in your local.conf, environment or other configuration file.

Summary: There was 1 ERROR message shown, returning a non-zero exit code.
````

**Describe the results you expected:**

The build to run.

**Additional information you deem important (e.g. issue happens only occasionally):**

````
$ cat conf/auto.conf 
MACHINE ?= "qemuriscv32"
#IMAGE_FEATURES += "tools-debug"
#IMAGE_FEATURES += "tools-tweaks"
#IMAGE_FEATURES += "dbg-pkgs"
# rootfs for debugging
#IMAGE_GEN_DEBUGFS = "1"
#IMAGE_FSTYPES_DEBUGFS = "tar.gz"
EXTRA_IMAGE_FEATURES_append = " ssh-server-dropbear"
EXTRA_IMAGE_FEATURES_append = " package-management"
PACKAGECONFIG_append_pn-qemu-native = " sdl"
PACKAGECONFIG_append_pn-nativesdk-qemu = " sdl"
USER_CLASSES ?= "buildstats buildhistory buildstats-summary image-mklibs image-prelink"

require conf/distro/include/no-static-libs.inc
require conf/distro/include/yocto-uninative.inc
require conf/distro/include/security_flags.inc

INHERIT += "uninative"

DISTRO_FEATURES_append = " largefile opengl ptest multiarch wayland pam  systemd "
DISTRO_FEATURES_BACKFILL_CONSIDERED += "sysvinit"
VIRTUAL-RUNTIME_init_manager = "systemd"
HOSTTOOLS_NONFATAL_append = " ssh"
````

if I change that qemuriscv32 back to qemuriscv64 it starts building, but I need a 32-bit setup to test QEMU.

[...]

PR #157, opened 2019-07-11
Add meta-python before meta-multimedia

palmer-dabbelt:
meta-multimedia depends on meta-python.  I'm not sure why it's
necessary to add all the dependent layers in order, but if I don't do
this I get a bunch of exceptions during setup.sh that result in no
meta-python in my bblayers.conf and thus a non-functional bitbake.

setup.sh was already explicitly adding meta-python, so this patch just
adds it before meta-multimedia.  That fixes the errors on my side.

Signed-off-by: Palmer Dabbelt <palmer@sifive.com>

### DIFF
--- a/setup.sh
+++ b/setup.sh
@@ -41,8 +41,8 @@ export BASH_SOURCE="openembedded-core/oe-init-build-env"
 # add the missing layers
 echo "Adding layers"
 bitbake-layers add-layer ../meta-openembedded/meta-oe
-bitbake-layers add-layer ../meta-openembedded/meta-multimedia
 bitbake-layers add-layer ../meta-openembedded/meta-python
+bitbake-layers add-layer ../meta-openembedded/meta-multimedia
 bitbake-layers add-layer ../meta-openembedded/meta-networking
 bitbake-layers add-layer ../meta-riscv
 


alistair23:
> **- : Short log / Statement of what needed to be changed.**
> 
> **-(Optional pointers to external resources, such as defect tracking)**
> 
> **-The intent of your change.**
> 
> **-(Optional, if it's not clear from above) how your change resolves the issues in the first part.**
> 
> **-Tag line(s) at the end.**
> 
> **-Signed-off-by: Random J Developer [random@developer.example.org](mailto:random@developer.example.org)**

Can you remove all this extra stuff?

alistair23:
This is the order I have, although I don't think it matters and no one else has complained so I don't see how this is needed. You might have something else wrong.

But the patch isn't going to break anything so if this fixes it for you it should be ok to merge.

palmer-dabbelt:
Ya, I don't understand what's going on.  I was pretty surprised this fixed the issue, as I would expect adding a layer to add layers dependent on it.  I haven't done any OE stuff before so I was just trying things at random.

pino-kim:
I reproduce issue #156 . This PR is related?

alistair23:
Ah ok, so it seems like something is wrong with the bblayers.conf. I'll try to reproduce this script setup and see what's going on

kraj:
order matters with `bitbake-layers add ...`

PR #161, opened 2019-07-24
freedom-u540: set UBOOT_DTB_LOADADDRESS

ricardosalveti:
Based on the default u-boot machine configuration, useful for fitImage
support (known address for loading the device-tree).


kraj:
@alistair23 can you review

alistair23:
This LGTM. I haven't tested it, but the address seems sane.

alistair23:
Plus this is only used in fitImages, so it won't break what we are currently using :)

PR #162, opened 2019-07-24
wic: add freedom-u540 wic for sota/ostree

ricardosalveti:
Wic file to be used with sota/ostree (meta-updater), as the source for
root needs to be provided by the otaimage plugin.

It can be maintained outside this layer, but having it maintained at the
official meta-riscv layer makes it a lot easier for users experimenting
with meta-updater.


ricardosalveti:
@alistair23 ^^

Following the suggestion from https://github.com/advancedtelematic/meta-updater/pull/565

PR #163, opened 2019-07-24
freedom-u540: prefer += instead of append for IMAGE_FSTYPES

ricardosalveti:
Append causes IMAGE_FSTYPES to contaminate initramfs-based images, as
even when forcing it to be empty at the recipe level, it always shows up
after parsing.

Using += allows the recipe to set another value without being influenced
by the machine configuration.

### DIFF
--- a/conf/machine/freedom-u540.conf
+++ b/conf/machine/freedom-u540.conf
@@ -28,7 +28,7 @@ SERIAL_CONSOLES = "115200;ttySIF0"
 
 MACHINE_EXTRA_RRECOMMENDS += " kernel-modules"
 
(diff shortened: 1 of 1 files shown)


ricardosalveti:
@alistair23 this should help fixing the issue you had.

alistair23:
This doesn't help the issue

ricardosalveti:
What is the error you get? I still need to empty EXTRA_IMAGEDEPENDS, but this is to avoid another circular dependency. You can check with this change that the wic image is not going to be in IMAGE_FSTYPES when building initramfs-ostree-image.

alistair23:
Ah, it was just the missing empty `EXTRA_IMAGEDEPENDS`

PR #165, opened 2019-07-25
freedom-u540: Use fitImage

alistair23:
Let's swap to using a fitImage by default. At the same time let's update
the u-boot boot commands to handle all the image types and support
loading a uEnv.txt file.


alistair23:
Thanks @kraj

PR #167, opened 2019-08-06
u-boot: Add OSTree uEnx.txt.in file

alistair23:
Add the uEnx.txt.in file to meta-riscv to allow users to build OSTree
setups for the HiFive Unleashed. This is based on the original version
from Ricardo Salveti (@ricardosalveti).

### DIFF
--- a/recipes-bsp/u-boot/u-boot_2019.07.bbappend
+++ b/recipes-bsp/u-boot/u-boot_2019.07.bbappend
@@ -25,6 +25,8 @@ SRC_URI_append_freedom-u540 = " \
             file://tftp-mmc-boot.txt \
            "
 
+SRC_URI_append_freedom-u540_sota = " file://uEnv.txt"
+
 DEPENDS_append_freedom-u540 = " u-boot-tools-native"
 
 # Overwrite this for your server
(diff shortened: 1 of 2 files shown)


recipes-bsp/u-boot/u-boot_2019.07.bbappend
```diff
@@ -39,6 +40,11 @@ do_configure_prepend_freedom-u540() {
     fi
 }
 
+do_compile_append_freedom-u540_sota() {
```
kraj: this uses `sota` additional override, it that intentional ?

alistair23: Yes, this is aimed at OSTree which requires sota

pino-kim: @alistair23 
OSTree is like a remote img server?

alistair23: No, it's an immutable rootFS. Kind of like git for your file system. https://ostree.readthedocs.io/en/latest/

kraj: > Yes, this is aimed at OSTree which requires sota

ok then would it make sense to add the .in file to SRC_URI also under same override ?

alistair23: Good point!

recipes-bsp/u-boot/u-boot_2019.07.bbappend
```diff
@@ -39,6 +41,11 @@ do_configure_prepend_freedom-u540() {
...
+    sed -e 's/@@KERNEL_BOOTCMD@@/${KERNEL_BOOTCMD}/' \
+        "${WORKDIR}/uEnv.txt.in" > uEnv.txt
```
ricardosalveti: This was only done for the recipe to be compatible with multiple targets, and not really required here.

If you just rename uEnv.txt.in to uEnv.txt and install that file it should already be enough for what you need.

kraj: I like the idea of substitution and have a single file to generate output

alistair23:
I pushed an updated version

PR #172, opened 2019-08-15
riscv: Enable SECURITY_CFLAGS and GLIBC PIE for rv32/rv64

kraj:
GCC PIE works fine

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -16,13 +16,3 @@ BBFILES_DYNAMIC += " \
 "
 
 LAYERSERIES_COMPAT_riscv-layer = "warrior"
-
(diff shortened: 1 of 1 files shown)


kraj:
@alistair23 here we go.

alistair23:
Cool, I've tested RV32 and don't see anything strange. I'll build the rest overnight and test tomorrow.

alistair23:
I don't see any breakages on anything else

kraj:
does it boot fine too ?

alistair23:
It does boot, but double checking the log I just noticed that u-boot doesn't detect a network on the HiFive Unleased. I'll see if I can figure out why

alistair23:
Ah, that's a u-boot issue.

All working :)

PR #175, opened 2019-08-28
Update glibc for the 5.2 headers

alistair23:
OE-Core has upgraded to the 5.2 kernel headers. This means that all RV32
software now requires a 64-bit time_t. Update the glibc fork to an
implementation that works in a 64-bit time_t. At the same time we can
revert the linux kernel patch.

This also requires fixed to Busybox to fix build failures.


alistair23:
This PR relies on the current OE-Core master-next branch (with the 5.2 kernel headers).

Can people please test this and let me know any issues that you see?

alistair23:
I'm marking this as ready to review. It shouldn't be merged into the 5.2 headers are merged into master.

alistair23:
This should be ready to merge now

PR #176, opened 2019-08-30
setup.sh: Do not delete build dir

kraj:
When building in a non CI env this all is not
needed as the user might accidentally end up losing all the build
artifacts that are not intendeed to. On CI systems its better to delete
it via extra step outside setup


pino-kim:
I remove that line when build meta-riscv. I need this commit.

PR #179, opened 2019-09-10
ltp: Remove RISC-V Patches

alistair23:
These no longer appear to be required, so let's remove them.


alistair23:
Yep, `TCLIBC="musl" MACHINE=qemuriscv64 bitbake ltp` builds.

I haven't tested 32-bit as musl doesn't support RV32.

kraj:
> Yep, `TCLIBC="musl" MACHINE=qemuriscv64 bitbake ltp` builds.
> 
> I haven't tested 32-bit as musl doesn't support RV32.

thanks

PR #187, opened 2019-11-12
openocd: Use the RISC-V fork

alistair23:
We currently need the RISC-V fork of OpenOCD.

### DIFF
--- a/recipes-devtools/openocd/openocd_%.bbappend
+++ b/recipes-devtools/openocd/openocd_%.bbappend
@@ -1,0 +1,8 @@
+SRC_URI = " \
+    git://github.com/riscv/riscv-openocd.git;protocol=http;branch=riscv;name=openocd \
+    git://repo.or.cz/r/git2cl.git;protocol=http;destsuffix=tools/git2cl;name=git2cl \
+    git://repo.or.cz/r/jimtcl.git;protocol=http;destsuffix=git/jimtcl;name=jimtcl \
+    git://repo.or.cz/r/libjaylink.git;protocol=http;destsuffix=git/src/jtag/drivers/libjaylink;name=libjaylink \
+"
+
+SRCREV_openocd = "f93ede5401c711e55d9852986aa399c0318efb22"


recipes-devtools/openocd/openocd_%.bbappend
```diff
@@ -0,0 +1,8 @@
+SRC_URI = " \
```
kraj: Make SRC_URI also  with overrides so we can ensure that meta-riscv can be used in multi-BSP scenarios

alistair23: I can't get the override to work as we it to apply to openocd-native. Any ideas?

kraj: lets define a separate recipe `openocd_riscv.bb` and then in layer.conf set 
```
PREFERRED_VERSION_openocd-native = "riscv"
PREFERRED_VERSION_openocd = "riscv"
```

alistair23: Works for me. Now the problem is that we don't have a conf file for qemuriscv64 in meta-riscv

kraj: yes, qemuriscv64 is assumed to be working without meta-riscv atm so adding this in meta-riscv makes it depend on this layer, layer can only help the machines, which it maintains, so this could ideally be done for oe-core but it wont be accepted. So questions I have is that

1. Can we use oe-core version for native ? if not then can we generate patches that can be applied to native version in core ?

alistair23: The native version is the important part. In this case I want to use OpenOCD on my native machine to debug a board.

I think the diff is fairly large, it won't be just a few simple patches.

It probably doesn't matter though. Users can use the `U540` machine to target 64-bit boards and `qemuriscv32` to target 32-bit ones.

alistair23: I just pushed an update that works for me.

It doesn't set anything for `qemuriscv64` machine, but I think that's ok.

kraj: Sounds good

PR #192, opened 2019-12-18
openocd: Enable remote bitbang

alistair23:
Enable remote bitbang to allow connecting OpenOCD to Verilator.

### DIFF
--- a/recipes-devtools/openocd/openocd_riscv.bb
+++ b/recipes-devtools/openocd/openocd_riscv.bb
@@ -8,3 +8,5 @@ SRC_URI = " \
 "
 
 SRCREV_openocd = "f93ede5401c711e55d9852986aa399c0318efb22"
+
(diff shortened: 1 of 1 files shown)


recipes-devtools/openocd/openocd_riscv.bb
```diff
@@ -8,3 +8,5 @@ SRC_URI = " \
 "
 
 SRCREV_openocd = "f93ede5401c711e55d9852986aa399c0318efb22"
+
+EXTRA_OECONF_append = " --enable-remote-bitbang"
```
kraj: this option should either be marked with `_riscv64` overrrides or sent to oe-core

alistair23: Yep, good point. I'll send it upstream.

PR #207, opened 2020-02-11
Add support for genering Xvisor images

alistair23:
Now that Xvisor is supported in [meta-virtulisation](https://git.yoctoproject.org/cgit/cgit.cgi/meta-virtualization/commit/?id=eb80f0c0ba0c8d088587eb2b1c6e7a561a4768ec) let's add support for the RISC-V specific things we need.


dynamic-layers/recipes-extended/xvisor/xvisor_%.bbappend
```diff
@@ -0,0 +1,67 @@
+XVISOR_PLAT_BASE = "`basename ${XVISOR_PLAT}`"
+
+do_compile[depends] += "virtual/kernel:do_deploy"
+
+do_compile_append() {
```
kraj: this append should be riscv specific I think

alistair23:
I have updated the PR

PR #237, opened 2020-08-14
openocd: Fix build with gcc10/-fno-common

kraj:


alistair23:
I really want to remove this OpenOCD fork. It seems like SiFive have no intention of upstreaming their work. Upstream OpenOCD does have RISC-V support, I might look into what we are still missing.

PR #241, opened 2020-08-26
linux-yocto : Support kernel 5.8 for riscv32

pino-kim:
Support kernel 5.8 on riscv architecture


pino-kim:
I check bootup kernel 5.8 on riscv32 with qemu.

I'm working on linux-mainline too.
Can you hold this PR, please?

alistair23:
This looks good. I'm hoping we can just support qemuriscv32 in the upstream linux-yocto recipe and drop this. But that's not going to happen for awhile, so this looks good in the meantime.

alistair23:
As for a linux-mainilne 5.8 I was thinking we should just drop support for linux-mainline. It uses the defconfig and the newer defconfigs have poor performance for RISC-V. linux-mainline also doesn't play well with other layers. Migrating over to linux-yocto seems like a better approach.

pino-kim:
@alistair23  
If am I correct only **freedom-u540** use linux-mainline current HEAD.
**I understand your comment** linux-yocto is a better approach to all three machines we support. Because not perform well in current linux-mainline 5.8. Am I right???

kraj:
> As for a linux-mainilne 5.8 I was thinking we should just drop support for linux-mainline. It uses the defconfig and the newer defconfigs have poor performance for RISC-V. linux-mainline also doesn't play well with other layers. Migrating over to linux-yocto seems like a better approach.

linux-yocto tooling has per BSP branches and metadata, which we have for qemu but not for freedom-u540, so if some one can validate this with linux-yocto we might be fine to use it. Or perhaps we can just use the upstream branch, I have not really tried it for freedom-u540

kraj:
> I check bootup kernel 5.8 on riscv32 with qemu.
> 
> I'm working on linux-mainline too.
> Can you hold this PR, please?

that can be independent pull, this in itself is fine, I have merged it.

pino-kim:
@kraj 
Thanks for all your reply. 
We can use both kernel linux-yocto and upstream. 
But I agree with @alistair23 suggestion too.
Ok, let's target linux-yocto to support **freedom-u540**
Thanks.

alistair23:
The problem is that post Linux 5.4 the defconfig for RISC-V has enabled a lot of debugging options by default. This results in a pretty significant slow down when running Linux. We could of course apply overrides to disable this, but that just covers up the problem.

Now that RISC-V has stabalised I think we should migrate away from using linux-mainline and instead swap to linux-yocto and use their config builder. This would bring us more inline with OE. For example the linux-mainline doesn't work with meta-agl as they don't have expect people to be using linux-mainline.

As Khem mentioned we would need to deal with the BSP and metadata, but it's probably worth figuring that out now anyway.

kraj:
I am ok if someone gets the BSP going.

PR #242, opened 2020-08-28
glibc: Update to use upstream

alistair23:
RISC-V 32-bit (RV32) support is now in upstream glibc!

This patch updates meta-riscv to use that, no more forks!


kraj:
Well done !! perseverance paid off

recipes-core/glibc/glibc_%.bbappend
```diff
@@ -1,12 +1,2 @@
-# Upstream Glibc does not yet have RISC-V 32-bit support
```
kraj, line 1: perhaps rename the bbappend to glibc_2.32.bbappend so it is disengaged once we upgrade to 2.33 in core, since we will have the support on oe-core then

alistair23: Good idea

PR #246, opened 2020-09-01
Use https URLs to fetch repositories

sorear:
The git:// URLs currently used provide no protection against MITM.

Signed-off-by: Stefan O'Rear <sorear@fastmail.com>


alistair23:
Do we not need the yocto one anymore?

kraj:
> Do we not need the yocto one anymore?

perhaps not but its there for convenience if someone wants to use poky, I will accept a removal patch

PR #253, opened 2020-10-16
Fix boost on rv32 and poppler on rv64

kraj:


recipes-support/boost/boost/riscv32__NR_futex.patch
```diff
@@ -11,3 +19,16 @@
  // Some Android NDKs (Google NDK and older Crystax.NET NDK versions) don't define SYS_futex
  #if defined(SYS_futex)
  #define BOOST_LOG_SYS_FUTEX SYS_futex
+--- a/boost/fiber/detail/futex.hpp
++++ b/boost/fiber/detail/futex.hpp
+@@ -7,6 +7,10 @@
+ #ifndef BOOST_FIBERS_DETAIL_FUTEX_H
+ #define BOOST_FIBERS_DETAIL_FUTEX_H
+ 
++#if !defined(__NR_futex) && defined(__riscv) && __riscv_xlen == 32
```
alistair23, line 28: We don't need the `__riscv` checks here, this isn't RISC-V specific, just 64-bit time_t on 32-bit specific.

recipes-support/boost/boost/riscv32__NR_futex.patch
```diff
@@ -1,3 +1,11 @@
+Define __NR_futex in terms of __NR_futex_time64 for rv32
```
alistair23, line 1: Could this go in OE instead? It isn't RISC-V specific.

recipes-support/boost/boost/riscv32__NR_futex.patch
kraj: other 32 bit architectures define `__NR_futex` so this is not needed for others.

recipes-support/boost/boost/riscv32__NR_futex.patch
kraj: yes it can perhaps more importantly it needs to be submitted upstream

recipes-support/boost/boost/riscv32__NR_futex.patch
alistair23: Other 32-bit architectures using 64-bit time_t don't. This includes Arc, but will be other architectures in the future.

kraj: right, I meant old 32bit architectures, so what is the right condition for detecting this case ? perhaps 
`#if !defined(__NR_futex) && defined(__NR_futex_time64)` might get the right condition

alistair23: Yep, that seems better

alistair23 (approved):

alistair23 (approved):

PR #259, opened 2021-01-01
musl: Add bits/reg.h for riscv32

kraj:


alistair23:
Can you add some details on why this is needed?

PR #273, opened 2021-03-16
OpenOCD: Bump SHA and enable jtag_vpi config option

alistair23:


dynamic-layers/openembedded-layer/recipes-devtools/openocd/openocd_riscv.bb
```diff
@@ -11,4 +11,6 @@ SRC_URI = " \
 
 SRCREV_openocd = "50a5971be27ff11061776afc8014bbff22870b7c"
 
+EXTRA_OECONF = "--enable-ftdi --disable-doxygen-html --disable-werror --enable-jtag_vpi"
```
kraj, line 14: should this be enabled in main recipe as well in meta-oe?

PR #284, opened 2021-05-10
linux-starfive: Switch back to starfive_vic7100_evb_sd_net_defconfig

kraj:
The real problem is that we can not boot kernel compiled with
-Os/gcc-11, once we disable CONFIG_CC_OPTIMIZE_FOR_SIZE, original
defconfig boots fine. Since its smaller in size and does not enable
ton of modules like the fedora config, lets switch back to this


alistair23 (approved):

alistair23:
Do you have steps to enable and setup WiFi?

kraj:
> Do you have steps to enable and setup WiFi?

it does not work yet :( I was just trying to ensure that we start putting kernel configs together

PR #285, opened 2021-05-12
linux-starfive: Enable stmmac.chain_mode and earlycon and add linux-mainline-dev kernel + wifi

kraj:


alistair23:
What does `stmmac.chain_mode` do?

pdp7:
From /Documentation/networking/stmmac.txt
> STMMAC supports DMA descriptor to operate both in dual buffer (RING)
> and linked-list(CHAINED) mode. In RING each descriptor points to two
> data buffer pointers whereas in CHAINED mode they point to only one data
> buffer pointer. RING mode is the default.
>  
> In CHAINED mode each descriptor will have pointer to next descriptor in
> the list, hence creating the explicit chaining in the descriptor itself,
> whereas such explicit chaining is not possible in RING mode.

kraj:
> What does `stmmac.chain_mode` do?

I was looking at driver parameters at https://www.kernel.org/doc/Documentation/networking/stmmac.txt
there are two modes `ring` and `chained` and `ring` is default, I just tried to use the non-default mode
there might be problems with ring mode in the driver who knows.

alistair23:
Does using chained fix something?

alistair23:
Ah, sorry. Just saw the commit message (why GitHub hides it I don't understand)

alistair23 (approved):

recipes-kernel/linux/linux-mainline-dev.bb
```diff
@@ -0,0 +1,40 @@
+SUMMARY = "BeagleV upstream dev kernel recipe"
```
alistair23, line 1: This isn't linux-mainline though

recipes-kernel/linux/linux-mainline-dev.bb
```diff
@@ -0,0 +1,40 @@
...
+
+LINUX_VERSION ?= "5.13.0"
+LINUX_VERSION_EXTENSION_append = "-starlight"
+
+KERNEL_VERSION_SANITY_SKIP = "1"
+SRCREV = "bda6ee4b4e84b2a68df0c0944b9b6bb836b72328"
+
+LIC_FILES_CHKSUM = "file://COPYING;md5=6bc538ed5bd9a7fc9398086aedcd7e46"
+
+PV = "${LINUX_VERSION}"
```
alistair23, line 18: Can we re-use our existing `recipes-kernel/linux/linux-mainline-common.inc`

alistair23 (changes requested):

alistair23 (approved):

conf/machine/beaglev-starlight-jh7100.conf
```diff
@@ -28,7 +28,7 @@ RISCV_SBI_FDT ?= "jh7100-starlight.dtb"
 
 SERIAL_CONSOLES = "115200;ttyS0"
 
-MACHINE_EXTRA_RRECOMMENDS += " kernel-modules"
+MACHINE_EXTRA_RRECOMMENDS += " kernel-modules linux-firmware-beaglev-bcm43430"
```
alistair23, line 31: Where does this come from?

kraj: hmm this comes from linux-starfive-dev recipe hmmm. We should switch to it as default I think

alistair23: Yeah, we probably should. Otherwise we should drop this as it fails to build by default.

PR #292, opened 2021-06-01
beaglev: opensbi: Use mainilne OpenSBI

alistair23:


recipes-bsp/u-boot/u-boot-starfive_v2021.04.bb
```diff
@@ -3,7 +3,7 @@ require recipes-bsp/u-boot/u-boot.inc
 
 FILESEXTRAPATHS_prepend := "${THISDIR}/files:"
 
-SRC_URI = "git://github.com/starfive-tech/u-boot.git;protocol=git;branch=Fedora_VIC_7100_2021.04 \
+SRC_URI = "git://github.com/atishp04/u-boot.git;protocol=git;branch=beagleV/Fedora_VIC_7100_2021.04 \
```
kraj, line 6: why should we switch to this fork ?

alistair23: Apparently that's the one to use: https://github.com/starfive-tech/opensbi/issues/4#issuecomment-852488314

kraj: OK and I hope this is short lived and eventually all these changes will flow into upstream u-boot ?

alistair23: That's my hope. It seems to only be two patches on top of the starfive branch, so hopefully we can switch to upstream or starfive soon.

lbmeng:
See the StarFive U-Boot changes to work with upstream OpenSBI (https://github.com/starfive-tech/u-boot/pull/17)

PR #295, opened 2021-06-07
Misc fixes for beagleV

kraj:


conf/machine/beaglev-starlight-jh7100.conf
```diff
@@ -30,7 +30,7 @@ SERIAL_CONSOLES = "115200;ttyS0"
 MACHINE_EXTRA_RRECOMMENDS += " kernel-modules linux-firmware-beaglev-bcm43430"
 
 IMAGE_FSTYPES += "wic.gz wic.bmap ext4"
-KERNEL_DEVICETREE ?= "starfive/jh7100-starlight.dtb"
+KERNEL_DEVICETREE ?= "starfive/jh7100-beaglev-starlight.dtb"
```
alistair23, line 33: I saw this as well. Won't this only work with the dev branch and break the other branch (which is currently the default)

kraj: yes @Esmil has confirmed that he will change it on 5.12 branch as well.

pdp7: There is not much value in the 5.10 branch ( https://github.com/starfive-tech/linux/tree/fedora-vic-7100_5.10.6) other than it is what Fu Wei (@tekkamanninja) is still using for the [Fedora image](https://github.com/starfive-tech/Fedora_on_StarFive) since Fedora 33 is stuck back on 5.10.  Akira (@mcd500) has actually been [repackaging the Fedora image to have the 5.13 kernel branch](https://github.com/mcd500/Fedora_on_StarFive) from @esmil.

People compiling their own systems (such as with meta-riscv) should be using the 5.13 kernel as it is the focus of all development and fixes: https://github.com/starfive-tech/linux/tree/esmil_starlight

Also, no need to reference @esmil's [own linux repo](https://github.com/esmil/linux/tree/starlight) as Emil can now push directly to starfive-tech/linux/esmil_starlight and it stays in sync with esmil/linux/starlight

kraj: @pdp7 thanks for updated info. I will add a patch to switch to https://github.com/starfive-tech/linux/tree/esmil_starlight
can you confirm that esmil plan's to actively maintain that branch instead of his own fork ?

pdp7: @alistair23 this discussion did make me realize that starfive-tech/linux should have a 5.13 kernel as the default branch. I have created https://github.com/starfive-tech/linux/issues/26 and will followup

alistair23 (approved):

PR #317, opened 2022-03-03
baremetal-riscv32nf: Fix typo in syntax override

angolini:
I believe there is a typo in this file, but I would like to double check.


kraj (approved):
lgtm Thanks @angolini

PR #324, opened 2022-03-24
linux-mainline: Bump to 5.17

alistair23:
Bump the linux-mainline to the 5.17 release. This means we are also
dropping the old SiFive U540 PCIe patches as they are no longer
maintained and the board is rarely used.


kraj:
I wonder if we can get `unmatched` board booted with this kernel.

alistair23:
I don't have one, so I can't test. But hopefully we can

kraj (approved):
lgtm

Issue #325, opened 2022-03-24
Weird build issue with meta-riscv


arunlee77:
**Description**

Trying to follow the README of meta-riscv i get the following error

while running

`. ./meta-riscv/setup.sh
`
```
Adding layers
++ bitbake-layers add-layer ../meta-openembedded/meta-oe
NOTE: Starting bitbake server...
/usr/lib/python3.6/importlib/_bootstrap.py:219: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds)
++ bitbake-layers add-layer ../meta-openembedded/meta-python
NOTE: Starting bitbake server...
/usr/lib/python3.6/importlib/_bootstrap.py:219: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds)
++ bitbake-layers add-layer ../meta-openembedded/meta-multimedia
NOTE: Starting bitbake server...
/usr/lib/python3.6/importlib/_bootstrap.py:219: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds)
++ bitbake-layers add-layer ../meta-openembedded/meta-networking
NOTE: Starting bitbake server...
/usr/lib/python3.6/importlib/_bootstrap.py:219: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds)
++ bitbake-layers add-layer ../meta-riscv
NOTE: Starting bitbake server...
/usr/lib/python3.6/importlib/_bootstrap.py:219: ImportWarning: can't resolve package from __spec__ or __package__, falling back on __name__ and __path__
  return f(*args, **kwds)

```
and then

`MACHINE=qemuriscv64 bitbake core-image-full-cmdline
`

```
MACHINE=qemuriscv64 bitbake core-image-full-cmdline
++ echo ---------------------------------------------------
---------------------------------------------------
++ echo ''

++ echo 'Buildable machine info'
Buildable machine info
++ echo ---------------------------------------------------
---------------------------------------------------
++ echo '* qemuriscv64: The 64-bit RISC-V machine'
* qemuriscv64: The 64-bit RISC-V machine
++ echo '* qemuriscv32: The 32-bit RISC-V machine'
* qemuriscv32: The 32-bit RISC-V machine
++ echo '* freedom-u540: The SiFive HiFive Unleashed board'
[...]

PR #326, opened 2022-03-24
conf: Drop using image-prelink

kraj:
image-prelink has been removed in latest yocto ( kirkstone + )
Fixes Issue #325


arunlee77 (commented):
Fixes the issue

PR #333, opened 2022-04-14
python3-pyruvate: Move to dynamic-layer for meta-python

tprrt:
Currently meta-riscv does not parse unless meta-python is present.

Moreover, it does not appear that anything explicitly depends on
python3-pyruvate in the layer.

Signed-off-by: Thomas Perrot <thomas.perrot@bootlin.com>


kraj (approved):
thanks for your contribution @tprrt

PR #348, opened 2022-07-13
libstd-rs: Remove upstreamed rv32 patch

kraj:


alistair23 (approved):

alistair23:
Content looks good, but the firmware patch should probably be separate. It isn't mentioned in the PR text

Issue #350, opened 2022-08-05
Unable to build images for VisionFive


dtaillard:
**Description**

The VisionFive machine configuration appears to be broken.

**Steps to reproduce the issue:**

Build the `core-image-base` image for the `visionfive` machine.

```bash
MACHINE=visionfive bitbake core-image-base
```

**Result:**

```
ERROR: u-boot-starfive-1_v2021.04-r0 do_populate_lic: QA Issue: u-boot-starfive: The LIC_FILES_CHKSUM does not match for file://Licenses/README;md5=2ca5f2c35c8cc335f0a19756634782f1
u-boot-starfive: The new md5 checksum is 5a7450c57ffe5ae63fd732446b988025
u-boot-starfive: Here is the selected license text:
vvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvvv
SPDX-License-Identifier: GPL-2.0
<omitted for brevity>
ERROR: u-boot-starfive-1_v2021.04-r0 do_populate_lic: Fatal QA errors were found, failing task.
ERROR: Logfile of failure stored in: <redacted>/riscv-yocto/build/tmp-glibc/work/visionfive-oe-linux/u-boot-starfive/1_v2021.04-r0/temp/log.do_populate_lic.141516
ERROR: Task (<redacted>/riscv-yocto/meta-riscv/recipes-bsp/u-boot/u-boot-starfive_v2021.04.bb:do_populate_lic) failed with exit code '1'
```

alistair23:
This is fixed by https://github.com/riscv/meta-riscv/pull/351

PR #351, opened 2022-08-08
u-boot: starfive_v2021.04: Override the u-boot LIC_FILES_CHKSUM

alistair23:
The LIC_FILES_CHKSUM md5sum changed with the latest mainline u-boot
release, but hasn't changed for the older starfive release. So manually
specify the original md5sum value.

Resolves: https://github.com/riscv/meta-riscv/issues/350


kraj (approved):
Lgtm

PR #371, opened 2022-11-04
ae350-ax45mp: Bumps to newer version

lyctw:
This patchset bumps following packages to the newer versions:

- OpenSBI v1.1 + fdt driver support
- U-boot v2022.10
- Linux 6.0.2

Linux kernel is hosted on AndesTech Github which includes ethernet, SD card, DMAC, RTC, WDT drivers support. OpenSBI is based on v1.1 with andes platfrom fdt driver, hence adding patches for U-boot to update the plmt and plicsw compatible strings and modify the IPI scheme.


recipes-bsp/opensbi/opensbi_%.bbappend
```diff
@@ -8,3 +8,6 @@ SRC_URI:append:nezha = " \
 "
 
 INSANE_SKIP:${PN}:nezha += "ldflags"
+
+# Support fdt drivers for AE350
+SRCREV:ae350-ax45mp = "22f38ee6c658a660083aa45c4ec6c72f66a17260"
```
kraj, line 13: can we bump it in OE-Core along side too? so we can remove this once in core ?

recipes-bsp/u-boot/files/ae350-ax45mp/0002-spl-Align-device-tree-blob-address-at-8-byte-boundar.patch
```diff
@@ -0,0 +1,38 @@
...
+
+Since libfdt v1.6.1, a new requirement on the device tree address via:
+
+  commit 5e735860c478 ("libfdt: Check for 8-byte address alignment in fdt_ro_probe_()")
+
+must be met that the device tree must be loaded in to memory at an
+8-byte aligned address.
+
+Upstream-Status: Pending
+Signed-off-by: Bin Meng <bmeng.cn@gmail.com>
```
kraj, line 14: what are plans for upstreaming it into u-boot ?

lyctw: Well.. there seems to be no solution on this [thread](https://patchwork.ozlabs.org/project/uboot/patch/20210712035231.26475-1-bmeng.cn@gmail.com/), but we did run into the misaligned issue without this patch.

recipes-bsp/opensbi/opensbi_%.bbappend
lyctw: OpenSBI v1.2 is not yet available.
:(

alistair23 (approved):

PR #376, opened 2022-12-23
Support HDMI output for Nezha board and updated Kernel and U-Boot as needed

cordlandwehr:
This patch set updates the Nezha board u-boot and kernel to support Kernel 6.1 together with the patch set by maeul, which enables HDMI graphics output. In order to do so:
- boot0 recipe could be removed because it is now obsolete according to https://linux-sunxi.org/Allwinner_Nezha
- WKS and nezha-allwinner-d1.conf are updated for more mainline boot scheme
- opensbi patches are dropped that were fully unused (using machine name "nezha" instead of "nezha-allwinner-d1" and according to my testing were not needed)
- introduced defconf file from previously used kernel fork and extended it to build sun4i driver, which is needed to use HDMI

I marked this PR as draft for now for two main reasons, where I would like to get feedback:
- is it okay to use the maeul branches? since I am new to this community I am not sure if they can be considered to be stable
- has anyone an idea why I cannot make the device booting when using FIT kernel images?


kraj:
> This patch set updates the Nezha board u-boot and kernel to support Kernel 6.1 together with the patch set by maeul, which enables HDMI graphics output. In order to do so:
> 
> * boot0 recipe could be removed because it is now obsolete according to https://linux-sunxi.org/Allwinner_Nezha
> * WKS and nezha-allwinner-d1.conf are updated for more mainline boot scheme
> * opensbi patches are dropped that were fully unused (using machine name "nezha" instead of "nezha-allwinner-d1" and according to my testing were not needed)
> * introduced defconf file from previously used kernel fork and extended it to build sun4i driver, which is needed to use HDMI
> 
> I marked this PR as draft for now for two main reasons, where I would like to get feedback:
> 
> * is it okay to use the maeul branches? since I am new to this community I am not sure if they can be considered to be stable

Thats Samuel H's fork, it should be fine to use it.

> * has anyone an idea why I cannot make the device booting when using FIT kernel images?

I think there are patches for oe-core in this area. Have you tried with latest master of poky/oe-core ?

cordlandwehr:
I am testing on top of the langdale branch. But after a check of recent changes on meta-oe/poky I do not see there any commit that seems to be related. Currently I am stuck (manual testing via uboot shell) in 

=> bootargs=earlycon=sbi initcall_debug=0 console=ttyS0,115200 loglevel=8 root=/dev/mmcblk0p2  init=/sbin/init
=> load mmc 0:1 ${ramdisk_addr_r} fitImage
5426596 bytes read in 899 ms (5.8 MiB/s)
=> bootm ${ramdisk_addr_r} - ${fdtcontroladdr}
## Loading kernel from FIT Image at 41c00000 ...
   Using 'conf-1' configuration
   Trying 'kernel-1' kernel subimage
     Description:  Linux kernel
     Type:         Kernel Image
     Compression:  gzip compressed
     Data Start:   0x41c00114
     Data Size:    5424911 Bytes = 5.2 MiB
     Architecture: RISC-V
     OS:           Linux
     Load Address: 0x40200000
     Entry Point:  0x40200000
     Hash algo:    sha256
     Hash value:   0275144df5b48a47a3a7f40e1c1545c3ca5d92fb4987890811928a95395c9988
   Verifying Hash Integrity ... sha256+ OK
## Flattened Device Tree blob at 7fd3e1c0
   Booting using the fdt blob at 0x7fd3e1c0
Working FDT set to 7fd3e1c0
   Uncompressing Kernel Image
Error: inflate() returned -5
gzip compressed: uncompress error -1

However, it cannot be the compression of the fit image itself, because also uncompressed FIT image yields the same problem, yet giving up with uncompress error -28.

cordlandwehr:
Nevermind, found the problem and will push updated patch series. The problem is that 0001-sun20i-set-CONFIG_SYS_BOOTM_LEN.patch is not correctly applied and uncompressing the Kernel runs out of memory.

recipes-kernel/linux/linux-nezha-dev.bb
```diff
@@ -14,16 +14,19 @@ KMETA = "kernel-meta"
 # directories that may be missing"
 SRC_URI = " \
         git://github.com/${FORK}/linux.git;name=machine;protocol=https;branch=${BRANCH} \
-        git://git.yoctoproject.org/yocto-kernel-cache;type=kmeta;name=meta;branch=yocto-5.15;destsuffix=${KMETA} \
-        file://0001-riscv-fix-build-with-binutils-2.38.patch \
+        git://git.yoctoproject.org/yocto-kernel-cache;type=kmeta;name=meta;branch=master;destsuffix=${KMETA} \
```
alistair23, line 17: We should use `BRANCH` here

alistair23 (approved):
Looks like a great cleanup! Thanks for this

recipes-kernel/linux/linux-nezha-dev.bb
cordlandwehr: Do you mean that I shall use a new variable for the KMETA branch here? Maybe nobranch=1 might also be a solution.

kraj: I think using revelant branch of yocto-kernel-cache is good to have see https://git.yoctoproject.org/yocto-kernel-cache/
but I think since this is a dev recipe I expect it to be latest kernel and hence using master might not be a bad idea.

cordlandwehr: OK, I am fine with either way, just need a decision what to use :)

PR #386, opened 2023-01-26
linux-nezha/defconfig: update led config name

twoerner:
This configuration parameter was renamed from R329 to A100 in v5 of this patch set (which is currently up to v7).

Signed-off-by: Trevor Woerner <twoerner@gmail.com>

**- <recipename>: Short log / Statement of what needed to be changed.**
  
**-(Optional pointers to external resources, such as defect tracking)**
  
**-The intent of your change.**
  
**-(Optional, if it's not clear from above) how your change resolves the
issues in the first part.**
  
**-Tag line(s) at the end.**

**-Signed-off-by: Random J Developer <random@developer.example.org>**

PR #387, opened 2023-01-26
linux-nezha-dev: point to branch with defconfig

twoerner:
One of the other branches of Samuel Holland's kernel tree for the nezha includes everything the current branch includes plus a defconfig. Using this branch will keep our defconfig inline with the author's changes.

Signed-off-by: Trevor Woerner <twoerner@gmail.com>


twoerner:
This could be an alternative to https://github.com/riscv/meta-riscv/pull/386
We don't need both that one and this, so whichever you prefer...

twoerner:
I wrote a test script too:
https://github.com/twoerner/rgb-led-test

PR #389, opened 2023-02-05
Add MangoPi MQ Pro board config

landgraf:
  rtl8723ds-mod: Added wifi driver 
  mangopi-mq-pro.conf: Added machine conf based on nezha-allwinner-d1.conf

   Add support of the board to meta-riscv and poky

**-Signed-off-by: Pavel Zhukov <pavel@zhukoff.net>

### DIFF
--- a/recipes-kernel/rtl8723ds-mod/rtl8723ds.bb
+++ b/recipes-kernel/rtl8723ds-mod/rtl8723ds.bb
@@ -1,0 +1,18 @@
+SUMMARY = "Linux driver for RTL8723DS"
+DESCRIPTION = "${SUMMARY}"
+LICENSE = "GPL-2.0-only"
+LIC_FILES_CHKSUM = "file://COPYING;md5=1f6f1c0be32491a0c8d2915607a28f36"
+
+inherit module
+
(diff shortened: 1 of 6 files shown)


recipes-kernel/rtl8723ds-mod/rtl8723ds.bb
```diff
@@ -0,0 +1,18 @@
+SUMMARY = "Linux driver for RTL8723DS"
+DESCRIPTION = "${SUMMARY}"
+LICENSE = "GPL-2.0-only"
```
alistair23, line 3: This seems to be supported upstream, do we really need this?

landgraf: If I understood correctly only bluetooth is supported upstream, not wifi. And Armbian uses this driver too https://github.com/armbian/build/blob/7bf3c297cb4a482468590c5f20dec2dd66ab2e29/lib/functions/compilation/patch/drivers_network.sh#L423 .   But I can be mistaken here...

alistair23: It looks like you are right. That's unfortunate

alistair23 (approved):

PR #403, opened 2023-04-08
VisionFive2 update for Starfive Release v2.11.5

cordlandwehr:
Update firmware and Kernel for VF2_v2.11.5 release. Also refresh MESA-PVR patches to remove fuzz in patch 65.


kraj (approved):
lgtm

PR #416, opened 2023-06-21
Update VisionFive2 SDK to 3.0.4

cordlandwehr:
This small update introduces two main changes:

- update binary blobs and VF2 specific libs to latest SDK (eg. fixes video acceleration, fixes red taint of display output)
- modifies GPU driver to be built as module, which fixes/workarounds a common race condition on Kernel loading when GPU binary blobs are not there then driver is initialized

### DIFF
--- a/conf/machine/JH7110.inc
+++ b/conf/machine/JH7110.inc
@@ -35,11 +35,13 @@ SERIAL_CONSOLES = "115200;ttyS0"
 
 MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
     kernel-modules \
+    linux-firmware-visionfive2-imggpu \
     vdec-module \
     venc-module \
     jpu-module \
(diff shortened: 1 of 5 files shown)


Fishwaldo:
This breaks Kirkstone based builds due to the new PVR userspace blobs having a dependency on newer versions of glibc (not at my PC right now so can't confirm the actual version). 

You could pick a commit to the soft_3rdparty repo just before they updated the blobs for gcc 12.2 rebuild and a single additional patch. (But obviously this is going to not be feasible as future releases come out) 

Also, just a FYI, the latest Debian snapshot has newer versions of things like Mesa/and PvR drivers (based on their changelog) but they have not released the source yet. Not sure if you want to do all this work if another drop is imminent?

kraj:
@Fishwaldo I am willing to create a tag if that helps, but `master` branch is not intended to be used with `kirkstone` release.  at this time, all works well with master branch

PR #429, opened 2023-09-07
Fix 'Directory not empty' error for rvisionfive2-pvr-graphics_1.19

OOHehir:
**-rvisionfive2-pvr-graphics_1.19.bb : Fix 'Directory not empty' error**
    
**```rmdir```fails as folder contains a subdirectory (systmed). This PR fixes this**
  
**Based on a build using this [KAS setup](https://github.com/OOHehir/build-riscv/blob/main/poky-visionfive2.yml)**
  
**-Signed-off-by: Owen O'Hehir <oo.hehir@gmail.com>**

### DIFF
--- a/recipes-core/musl/musl/0001-riscv32-fix-inconsistent-ucontext_t-struct-tag.patch
+++ b/recipes-core/musl/musl/0001-riscv32-fix-inconsistent-ucontext_t-struct-tag.patch
@@ -7,6 +7,7 @@ ucontext.h depends on the internal struct tag name for namespacing
 reasons, and the intent was always for it to be consistent across
 archs anyway.
 
+Upstream-Status: Pending
 Signed-off-by: Khem Raj <raj.khem@gmail.com>
 ---
  arch/riscv32/bits/signal.h | 4 ++--
--- a/recipes-core/musl/musl/0007-Emulate-wait4-using-waitid.patch
+++ b/recipes-core/musl/musl/0007-Emulate-wait4-using-waitid.patch
@@ -17,6 +17,9 @@ a small cost in code size.
 
 wait4 is historically not a cancellation point in musl; we retain that
 since we need the non-cancellable version of __wait4 anyway.
+
+Upstream-Status: Pending
+Signed-off-by: Khem Raj <raj.khem@gmail.com>
 ---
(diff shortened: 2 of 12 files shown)


recipes-graphics/drivers/visionfive2-pvr-graphics_1.19.bb
```diff
@@ -40,7 +40,8 @@ do_install () {
     # let vulkan-loader from core layer provide libvulkan
     rm -rf ${D}${libdir}/libvulkan*.so* ${D}${libdir}/pkgconfig/vulkan.pc
     # provided via separate arch-independent firmware package
-    rm -rf ${D}/lib/firmware
+    # remove all firmware & systemd
+    rm -rf ${D}/lib/*
```
kraj, line 44: can you add more info about the systemd files we are deleting ? Perhaps they can be used on systemd based distros and only deleted for non-systemd distro settings.

OOHehir:
```
lib
└── systemd
    └── system
        └── rc.pvr.service
```

& contents of rc.pvr.service
```
[Unit]
Description=PowerVR consumer services

[Service]
ExecStart=/usr/bin/rc.pvr start
ExecStop=/usr/bin/rc.pvr stop
Type=oneshot
RemainAfterExit=yes

[Install]
WantedBy=multi-user.target
```

kraj:
> ```
> lib
> └── systemd
>     └── system
>         └── rc.pvr.service
> ```
> 
> & contents of rc.pvr.service
> 
> ```
> [Unit]
> Description=PowerVR consumer services
> 
> [Service]
> ExecStart=/usr/bin/rc.pvr start
> ExecStop=/usr/bin/rc.pvr stop
> Type=oneshot
> RemainAfterExit=yes
> 
> [Install]
> WantedBy=multi-user.target
> ```

yeah. this file will be needed for systemd based distros. So I would suggest to delete it conditionally only when systemd is not in DISTRO_FEATURES

OOHehir:
OK, I've updated the PR. I've tested with & without systemd, seems to be satisfactory.

kraj:
> OK, I've updated the PR. I've tested with & without systemd, seems to be satisfactory.

thanks this looks ok now. Please squash both commits into one and reupload the PR

PR #431, opened 2023-09-10
Fix directory not empty error when systemd folder present

OOHehir:
As previously [discussed](https://github.com/riscv/meta-riscv/pull/429)

**- visionfive2-pvr-graphics: Fix 'directory not empty' error when systemd folder present**
  
**-[Affected recipe](https://github.com/riscv/meta-riscv/blob/master/recipes-graphics/drivers/visionfive2-pvr-graphics_1.19.bb)**
  
**-Fix break in build**

**-Signed-off-by: Owen O'Hehir <oo.hehir@gmail.com>**


kraj (approved):
lgtm

PR #437, opened 2023-10-09
milkv-duo: Add initial support for Milk-V Duo board

xyq1113723547:
Milk-V Duo is an ultra-compact dual-core RISC-V development board from Sophgo using CVITEK CV1800B SOC.   

For more details please visit [milkv.io/duo](milkv.io/duo)

The Linux kernel, U-Boot, OpenSBI are open-sourced and placed in the official [SDK](github.com/milkv-duo/duo-buildroot-sdk), but not up-streamed yet.  

The required  firmware fip.bin in boot partition can compile from source in SDK, but it requires two custom down-stream toolchains and prebuilt object files in the SDK, so it's hard to integrate here.

kernel defconfig seems a bit redundant here, but it's not in the forked kernel source and official documentation says it's  indeed out of the kernel source tree [here](https://doc.sophgo.com/cvitek-develop-docs/master/docs_latest_release/CV180x_CV181x/en/01.software/OSDRV/U-boot_Porting_Development_Guide/build/html/2_Function_Overview.html).

Auto-generated .its in yocto differs from the specs that the firmware requires and thus can not boot. Therefore, it's better to provide multi.its, compile required cv1800b_milkv_duo_sd.dtb from my modified kernel version and generated the required boot.sd(kernel.itb) in boot partition by custom commands.


recipes-kernel/linux/linux-milkv-duo-dev/cvitek_cv1800b_milkv_duo_sd_defconfig
kraj, line 1: is this file generated with `make savedefconfig` ? if not it might be better to save the output of this command as defconfig, it will be small and able to generate correct .config during compile process eventually.

recipes-kernel/linux/linux-milkv-duo-dev.bb
```diff
@@ -0,0 +1,47 @@
...
+
+inherit kernel
+require recipes-kernel/linux/linux-yocto.inc
+
+DEPENDS = "u-boot-mkimage-native dtc-native"
+
+BRANCH ?= "main"
+FORK ?= "xyq1113723547"
+
+#  git://git.yoctoproject.org/yocto-kernel-cache;type=kmeta;name=meta;branch=master;destsuffix=${KMETA} \
```
kraj, line 11: I think it will be better to use 5.10 branch of meta tree as well. see https://git.yoctoproject.org/yocto-kernel-cache/log/?h=yocto-5.10

recipes-kernel/linux/linux-milkv-duo-dev.bb
```diff
@@ -0,0 +1,47 @@
+SUMMARY = "Milk-V Duo dev kernel recipe"
+
+inherit kernel
+require recipes-kernel/linux/linux-yocto.inc
+
+DEPENDS = "u-boot-mkimage-native dtc-native"
+
+BRANCH ?= "main"
+FORK ?= "xyq1113723547"
```
kraj, line 9: this seems your local fork, what the plan to upstream it ?

recipes-kernel/linux/linux-milkv-duo-dev.bb
xyq1113723547: Make sense.

recipes-kernel/linux/linux-milkv-duo-dev.bb
xyq1113723547: Personally I have no plan to upstream my fork. Sophgo is hiring for people who can upstream their kernel. Another team is also trying to do kernel upstream here: https://github.com/duosdk/linux/tree/duosdk. I provide my fork here because the kernel in the official SDK can not be built by current upstream toolchain because of Kconfig and Makefiles. Also, dtb generation is out of tree, so is not suitable in yocto's workflow right now.

kraj: OK as long as your fork is kept alive we can think about using it

recipes-kernel/linux/linux-milkv-duo-dev/multi.its
```diff
@@ -0,0 +1,56 @@
...
+			load = <0x0 0x80200000>;
+			entry = <0x0 0x80200000>;
+			hash-2 {
+				algo = "crc32";
+			};
+		};
+
+
+		/*FDT*/
+
```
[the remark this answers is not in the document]
kraj: I think using yocto kernel tooling would be an improvement for sure. It is not a whole lot of effort for upstream kernels. e.g. see 
https://github.com/akuster/meta-odroid/blob/master/recipes-kernel/linux/linux-hardkernel_5.10.bb

for an example

xyq1113723547:
The kernel upstream pending patches for Milkv Duo is here https://lore.kernel.org/lkml/20230930123937.1551-1-jszhang@kernel.org/T/#t. I guess I shall wait and change my PR when it is upstreamed.

dlan17:
> The kernel upstream pending patches for Milkv Duo is here https://lore.kernel.org/lkml/20230930123937.1551-1-jszhang@kernel.org/T/#t. I guess I shall wait and change my PR when it is upstreamed.

No, please don't wait for upstream kernel, it's totally a different story, it would take long time for the mainline kernel to get at a production level..

I'd suggest you trying to provide two kernel options, one for vendor kernel with full features, one for mainline kernel with clean code but with basic functionality (currently only uart driver works)..

recipes-kernel/linux/linux-milkv-duo-dev.bb
dlan17: > I think it will be better to use 5.10 branch of meta tree as well. see https://git.yoctoproject.org/yocto-kernel-cache/log/?h=yocto-5.10

em,  my second thought, the above comment left there only makes people confused, the milkv duo doesn't work with yocto-kernel-cache's yocto-5.10 branch, right? as it requires quite heavy extra patches.. so we should have that comment removed

xyq1113723547:
I clean the defconfig, multi.its and linux-milkv-duo-dev.bb as @dlan17 suggested. 
The `cvitek_cv1800b_milkv_duo_sd_defconfig` is now smaller, manually cleaned and verified, and enable IPv6 and Multicast by default. 
I also read the defconfig generated by `make saveconfig`, which is very similar, but contains more lines compare to this one.

PR #474, opened 2024-05-09
visionfive2-pvr-graphics: fix installation with SysV init

### LINKED ISSUE #473: visionfive2-pvr-graphics: failing installation with SysV init
**Description**

When `systemd` is not in `DISTRO_FEATURES`, visionfive2-pvr-graphics fails to install the corresponding init scripts. Due to an incorrect if condition, the init system check is always evaluated as systemd.

In line 35 `if [ ${@bb.utils.contains('DISTRO_FEATURES', 'systemd', 'true', 'false', d)} ]; then` - though the string returned is correct, the `[]` test always evaluates it as a `true` value, as it checks only string length.

Basically this will always evaluate as true, and echoes:
```
if [ false ]; then
  echo false
fi
```

The correct version would be
```
if false; then
  echo false
fi
```
which evaluates as it is intended to be.

OldManYellsAtCloud:
Fix the installation of this recipe when using SysV init system by fixig the "if" condition that checks the presence of systemd in DISTRO_FEATURES - beforehand it was always evaluated as "true".

Beside this the required folders are also created when they should be: in case of SysV init the /usr/bin folder stays empty (and bitbake complains), so it is only created when systemd is used.

Similarly, /etc/init.d is used only by SysV init system, but the recipe already handles it when systemd is used. However if the folder is not installed explicitly, the do_rootfs task fails.

Fixes #473

### DIFF
--- a/recipes-graphics/drivers/visionfive2-pvr-graphics_1.19.bb
+++ b/recipes-graphics/drivers/visionfive2-pvr-graphics_1.19.bb
@@ -23,7 +23,7 @@ PACKAGES += " \
 do_install () {
     tar xz --no-same-owner -f ${S}/IMG_GPU/out/${IMG_GPU_POWERVR_VERSION}.tar.gz -C ${D}
     mv ${D}/${IMG_GPU_POWERVR_VERSION}/target/* ${D}
-    install -d ${D}${includedir} ${D}${bindir}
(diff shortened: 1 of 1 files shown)


kraj (approved):
lgtm

PR #482, opened 2024-07-15
linux-milkv-duo[-dev]: replace WORKDIR with UNPACKDIR in do_deploy

hiagofranco:
Trying to bitbake milkv-duo machine triggers the following error:

| cp: cannot stat 'riscv-yocto/build/tmp/work/milkv_duo-poky-linux/
linux-milkv-duo-dev/1.0/multi.its': No such file or directory

| cp: cannot stat 'riscv-yocto/build/tmp/work/milkv_duo-poky-linux/ linux-milkv-duo/6.8++gitAUTOINC+b95f2066a9/multi.its': No such file or directory

This is caused by multi.its file being in UNPACKDIR instead of WORKDIR, following the recent upstream changes in OE.


kraj (approved):
lgtm thanks for providing fix.

PR #492, opened 2024-09-21
Use scarthgap branch in oe.yml (fixes scarthgap build)

kotylamichal:
It will be better to use specific release branch for oe/poky layers instead of master, because currently it throws layer compatibility error. It is similar to oe.yml in meta-riscv kirkstone branch, where branch kirkstone instead of master is used for oe/poky.

At this time, master in poky repo is compatible only with mickledore, scarthgap and nanbield.


kraj (approved):
seems good to me.

Issue #268, opened 2021-02-12
kernel modules fail to insert on qemuriscv32


kraj:
build a simple image which has some kernel modules also part of image and then it fails to insert the module.
see the below steps.

```
root@qemuriscv32:~# lsmod
Module                  Size  Used by
root@qemuriscv32:~# modinfo /lib/modules/5.10.12-yocto-standard/kernel/fs/binfmt_misc.ko
filename:       /lib/modules/5.10.12-yocto-standard/kernel/fs/binfmt_misc.ko
license:        GPL
alias:          fs-binfmt_misc
depends:        
intree:         Y
name:           binfmt_misc
vermagic:       5.10.12-yocto-standard SMP preempt mod_unload riscv
root@qemuriscv32:~# modprobe binfmt_misc
modprobe: ERROR: could not insert 'binfmt_misc': Invalid argument
root@qemuriscv32:~# uname -a 
Linux qemuriscv32 5.10.12-yocto-standard #1 SMP PREEMPT Thu Feb 11 20:16:28 UTC 2021 riscv32 riscv32 riscv32 GNU/Linux
root@qemuriscv32:~# zcat /proc/config.gz  | grep MODULE
CONFIG_MODULE_SECTIONS=y
CONFIG_MODULES_USE_ELF_RELA=y
CONFIG_MODULES=y
# CONFIG_MODULE_FORCE_LOAD is not set
CONFIG_MODULE_UNLOAD=y
# CONFIG_MODULE_FORCE_UNLOAD is not set
# CONFIG_MODULE_SRCVERSION_ALL is not set
# CONFIG_MODULE_SIG is not set
# CONFIG_MODULE_COMPRESS is not set
# CONFIG_MODULE_ALLOW_MISSING_NAMESPACE_IMPORTS is not set
CONFIG_MODULES_TREE_LOOKUP=y
root@qemuriscv32:~# 
```

pino-kim:
Hi, @kraj long time no see.
Is this issue still alive?

kraj:
yes it still fails

```
root@qemuriscv32:~# modprobe nfsd
[  141.976155] nfsd: target 00000000a06c442c can not be addressed by the 32-bit offset from PC = e173c030
modprobe: ERROR: could not insert 'nfsd': Invalid argument


root@qemuriscv32:~# uname -a
[...]

PR #500, opened 2024-11-01
Fix kernel modules insertion issue on qemuriscv32

vishwamartur:
Related to #268

Update kernel configuration and version for qemuriscv32

* Update `KBUILD_DEFCONFIG:qemuriscv32` to use `rv32_defconfig` in `linux-mainline-common.inc`.
* Remove `KERNEL_FEATURES:remove` entries for `features/debug/printk.scc` and `features/kernel-sample/kernel-sample.scc` in `linux-mainline-common.inc`.
* Update `SRC_URI` to include the new kernel version `5.13.5` in `linux-mainline_6.2.bb`.
* Update `LINUX_VERSION` to `5.13.5` in `linux-mainline_6.2.bb`.
* Add `qemuriscv32.conf` with updated `KERNEL_IMAGETYPE` to `Image` and `KERNEL_DEFCONFIG` to `rv32_defconfig`.

### DIFF
--- a/conf/machine/qemuriscv32.conf
+++ b/conf/machine/qemuriscv32.conf
@@ -1,0 +1,15 @@
+#@TYPE: Machine
+#@NAME: qemuriscv32
+#@DESCRIPTION: QEMU RISC-V 32-bit machine
+
+require conf/machine/include/qemu.inc
+require conf/machine/include/riscv-common.inc
+
+KERNEL_IMAGETYPE = "Image"
+KERNEL_DEFCONFIG = "rv32_defconfig"
+
+QB_SYSTEM_NAME = "qemu-system-riscv32"
+QB_MACHINE = "-machine virt"
+QB_CPU = "-cpu rv32"
+QB_MEM = "-m 2048"
+QB_KERNEL_CMDLINE_APPEND = "console=ttyS0"
--- a/recipes-kernel/linux/linux-mainline-common.inc
+++ b/recipes-kernel/linux/linux-mainline-common.inc
@@ -17,6 +17,3 @@ KBUILD_DEFCONFIG:qemuriscv64 = "defconfig"
 KBUILD_DEFCONFIG:freedom-u540 = "defconfig"
 
 COMPATIBLE_MACHINE = "(qemuriscv32|qemuriscv64|freedom-u540)"
-
-KERNEL_FEATURES:remove = "features/debug/printk.scc"
-KERNEL_FEATURES:remove = "features/kernel-sample/kernel-sample.scc"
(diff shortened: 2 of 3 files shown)


conf/machine/qemuriscv32.conf
```diff
@@ -0,0 +1,15 @@
+#@TYPE: Machine
+#@NAME: qemuriscv32
+#@DESCRIPTION: QEMU RISC-V 32-bit machine
```
alistair23, line 3: This board is upstream: https://github.com/openembedded/openembedded-core/blob/master/meta/conf/machine/qemuriscv32.conf

Any changes should be sent there

recipes-kernel/linux/linux-mainline_6.2.bb
```diff
@@ -1,10 +1,10 @@
 require recipes-kernel/linux/linux-mainline-common.inc
 
-LINUX_VERSION ?= "6.2+"
+LINUX_VERSION ?= "5.13.5"
```
alistair23, line 3: This looks wrong, it's the 6.2 kernel

recipes-kernel/linux/linux-mainline-common.inc
```diff
@@ -17,6 +17,3 @@ KBUILD_DEFCONFIG:qemuriscv64 = "defconfig"
 KBUILD_DEFCONFIG:freedom-u540 = "defconfig"
 
 COMPATIBLE_MACHINE = "(qemuriscv32|qemuriscv64|freedom-u540)"
-
-KERNEL_FEATURES:remove = "features/debug/printk.scc"
-KERNEL_FEATURES:remove = "features/kernel-sample/kernel-sample.scc"
```
alistair23, line 22: This removes it for everything, this isn't the right way

alistair23 (changes requested):

conf/machine/qemuriscv32.conf
vishwamartur: Thank you, @alistair23, for pointing that out! I’ll send any updates to the upstream repository, as suggested.

recipes-kernel/linux/linux-mainline_6.2.bb
vishwamartur: Thanks for catching that, @alistair23  I’ll update it to ensure it reflects the correct 6.2 kernel version.

recipes-kernel/linux/linux-mainline-common.inc
vishwamartur: Thanks for the feedback, @alistair23  I see your point. I’ll adjust it to target only the specific configurations where removing these features is necessary.

PR #507, opened 2024-12-27
Force the kernel git fecther protocol to be 'https'

gerryqd:
By default, the git fetcher protocol will be 'git', this will cause issue when you behind a firewall.

Change them to 'https' will help to resolve these issues.

issues in the first part.**
  
**-Tag line(s) at the end.**


kraj (approved):
lgtm

PR #518, opened 2025-04-07
[Styhead][U-Boot][AllWinnerD1] Fix folder for boot.scr.uimg

MartinHerren:
On the `Styhead` branch, during `u-boot-allwinnerd1`'s configure task, the file `boot.scr.uimg` gets installed in the `sources-unpack/` folder while the install/deploy tasks expect it in the `build/` folder, thus the build fails

This MR fixes the configure role to generate this file in the `build/` folder.

Tested generating both `mangopi-mq-pro` and `nezha` images that previously failed.

Attached full boot log on a mangopi-mq-pro:
[mangopi-mq-pro_fullboot_styhead.log](https://github.com/user-attachments/files/19639346/mangopi-mq-pro_fullboot_styhead.log)

Same fix might also be required on some other machines.


kraj (approved):
lgtm

MartinHerren:
For info, the same fix is also necessary on master. There is also an additional fix needed there, progressing well on it.
I'll open the MR for master as soon as i can.

MartinHerren:
Master fix PR opened at https://github.com/riscv/meta-riscv/pull/522

PR #522, opened 2025-04-11
[U-Boot][AllWinnerD1] Fix build of u-boot

MartinHerren:
During u-boot-allwinnerd1's configure task, the file boot.scr.uimg gets installed in the sources-unpack/ folder while the install/deploy tasks expect it in the build/ folder.

In addition the newer Python version is not compatible anymore with this u-boot version. I added a patch back ported from upstream u-boot that fixes this.

I will try to open a MR with this back ported patch to the u-boot fork used for AllWinner D1. Doesn't seem to be actively maintained, still worth a try

Tested generating and running on a mangopi-mq-pro, both against poki's master and walnascar branches.

Attached full boot log on a mangopi-mq-pro:
[mangopi-mq-pro_fullboot_master.log](https://github.com/user-attachments/files/19702327/mangopi-mq-pro_fullboot_master.log)

I'm still learning Yocto, so not sure about the formatting of the patch file. I tried to follow https://docs.yoctoproject.org/contributor-guide/recipe-style-guide.html#patch-upstream-status and put both `Upstream-Status` info for the u-boot fork and the official one.

### DIFF
--- a/recipes-bsp/u-boot/u-boot-allwinnerd1.bb
+++ b/recipes-bsp/u-boot/u-boot-allwinnerd1.bb
@@ -12,6 +12,7 @@ SRC_URI = " \
     file://tftp-mmc-boot.txt \
     file://uEnv-nezha.txt \
     file://0001-sun20i-set-CONFIG_SYS_BOOTM_LEN.patch \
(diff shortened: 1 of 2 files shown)


kraj (approved):
lgtm

PR #524, opened 2025-04-14
Fix beaglev support

### LINKED ISSUE #521: linux-beaglev-dev: wrong kernel version fetch
yocto always fetches the latest Linux kernel regardless of the "SRCREV[kernel]" value.
The recipe in linux-beaglev-dev.bb needs to be fixed.

[linux-beaglev-dev.bb](https://github.com/riscv/meta-riscv/blob/2085da7/recipes-kernel/linux/linux-beaglev-dev.bb):

```
SRCREV[kernel] = "ffd294d346d185b70e28b1a28abe367bbfe53c04"
SRCREV[dts] = "62943572e043726f7357d8f36cfbc5dfcba7ed3b"

BRANCH_kernel = "master"
BRANCH_dts = "v6.13.x"
LINUX_VERSION ?= "6.13.0"

SRC_URI = " \
    git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git;protocol=https;branch=${BRANCH_kernel};name=kernel \
    git://git.beagleboard.org/beagleboard/BeagleBoard-DeviceTrees.git;protocol=https;subdir=beaglev_devicetrees;branch=${BRANCH_dts};name=dts \
    file://defconfig \
    file://extlinux.conf \
```

kanakshilledar:
Fixes #521 
Also improves the overall support for this board.
Upgrade the kernel to v6.14


recipes-kernel/linux/linux-beaglev-dev.bb
```diff
@@ -37,7 +35,7 @@ do_configure:prepend() {
 DEPENDS += "e2fsprogs-native"
 
 # package a separate partition boot.ext4 that can be flashed via fastboot to partition boot
-do_beaglev_img_deploy() {
+do__deploy:append() {
```
OldManYellsAtCloud, line 38: This looks like a typo - there seem to be two underscores in `do_deploy` instead of one.

kanakshilledar: fixed it!
thanks for indicating :)

kanakshilledar:
Yes! I noticed this bug and fixing it. It turns out that `do_deploy:append()` wont be the solution. We will have to stick to the custom functions only. Like the earlier ones, but again it causes to not run automatically.

kanakshilledar:
The `u-boot-beaglev-ahead.bb` recipe doesn't run automatically. We have to trigger it automatically.

kraj (approved):
LGTM

OldManYellsAtCloud:
>  Like the earlier ones, but again it causes to not run automatically.

Another alternative you could test: have a custom function, but don't make it a task, rather a postfunc for `do_bundle_initramfs` (or whatever task generates the missing files). That way it will be executed every time when a new kernel is generated:
```
do_bundle_initramfs[postfuncs] += ";do_beaglev_img_deploy"
```

You will need to add `opensbi:do_deploy` as a dependency for the appropriate task too.

kanakshilledar:
> Another alternative you could test: have a custom function, but don't make it a task, rather a postfunc for `do_bundle_initramfs` (or whatever task generates the missing files). That way it will be executed every time when a new kernel is generated:

I tried this suggestion, but it was not building automatically.

OldManYellsAtCloud (approved):

PR #536, opened 2025-06-20
layer.conf: Update to whinlatter release series

kraj:
Since there are currently multiple breaking changes, update to a new release series for the next release so layers can show compatibility correctly.

issues in the first part.**
  
**-Tag line(s) at the end.**

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -23,7 +23,7 @@ BBFILES_DYNAMIC += " \
     \
 "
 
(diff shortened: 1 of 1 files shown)


alistair23 (approved):

OldManYellsAtCloud:
I think on its own this will break, without removing the `S = "${WORKDIR}/git"` (&co) assignments from the recipes.

PR #546, opened 2025-08-26
visionfive2-firmware: Point to strlcat.eu mirror

kraj:
starfive repo is not able to allow LFS fetches

This repository exceeded its LFS budget. The account responsible for the budget should increase it to restore access.

Until it is fixed in github.com/starfive-tech/soft_3rdpart.git Use a mirror

https://forum.rvspace.org/t/quota-problem-for-github-com-starfive-tech-soft-3rdpart/4597?utm_source=chatgpt.com


kraj:
@cordlandwehr the problem is back sadly :(

PR #556, opened 2025-10-15
conf/machine/orangepi-rv2.conf: weak assignment for KERNEL_DEVICETREE

michaelopdenacker:
Use weak assignment for KERNEL_DEVICETREE as in the other machine definitions in this layer. This makes it possible to set a custom or modified device tree in conf/local.conf.


kraj (approved):
LGTM

PR #567, opened 2025-11-23
opensbi: drop milk-v overrides

mischief:
with gcc 15 opensbi v1.4 no longer builds due to
https://github.com/riscv-software-src/opensbi/issues/365.

drop SRCREV/SRC_URI overrides for milkv-duo and use opensbi v1.7 from the core layer.

boot log showing 1.7 works in [opensbi17.txt](https://github.com/user-attachments/files/23693176/opensbi17.txt)


kraj (approved):
Thanks LGTM

PR #573, opened 2025-12-01
beaglev-ahead.conf: sparse images and add flashing script helper

lslrt:
- conf/machine/beaglev-ahead.conf: inherit image_types_sparse from meta-openembedded layer so the rootfs can be converted to a sparse image otherwise fastboot will likely fail to flash it.

- recipes-kernel/linux/linux-beaglev-dev.bb: sparse the boot.ext4 image to remove the 'fastboot flash boot boot.ext4' command generate a warning message "Invalid sparse file format at header magic"

- recipes-support/flashing-scripts/flashing-script_1.0.bb: introducing a new "deploy only recipe" that deploy a flash.sh utility script to deploydir/. This may help end user to flash the device.

Signed-off-by: Leo Sartre <leo@missingno.tech>

### DIFF
--- a/conf/layer.conf
+++ b/conf/layer.conf
@@ -23,6 +23,8 @@ BBFILES_DYNAMIC += " \
     \
 "
 
+LAYERDEPENDS_riscv-layer = "core openembedded-layer"
+
 LAYERSERIES_COMPAT_riscv-layer = "wrynose"
 
 LICENSE_PATH += "${LAYERDIR}/licenses"
--- a/.github/workflows/check-layer.yml
+++ b/.github/workflows/check-layer.yml
@@ -43,6 +43,7 @@ jobs:
         run: |
           git clone --depth 1 --branch master https://github.com/openembedded/openembedded-core.git
           git clone --depth 1 --branch master https://github.com/openembedded/bitbake.git openembedded-core/bitbake
+          git clone --depth 1 --branch master https://github.com/openembedded/meta-openembedded.git
           chown -R builduser:builduser $GITHUB_WORKSPACE
 
       - name: Run yocto-check-layer-wrapper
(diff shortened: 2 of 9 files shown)


kraj:
Please rebase this PR on latest master and also address the feedback.

lslrt:
Hey @kraj , I've rebased my work.
During the rebase, I've updated a bit the way the boot.ext4 sparse image is created using `ext2simg_android`, that make me realize that the dev kernel recipe for beaglev-ahead would really deserve a cleanup.
Generating the .ext4 image from kernel do_deploy task is not super elegant and it would be better to rely on wic to do that, with a clean image we could also remove this hack of sparsing the .ext4 file manually and rather rely on `image_types_sparse`.
That would then need a way to make this new boot.ext4 image linked as a dependencies to every image that are built for beaglev-ahead.... maybe using EXTRA_IMAGEDEPENDS in machine conf with a python inline hack to cut circular dependency?
I do not have much time to look at it now, shall we go with this first simple Pull request and open an issue to cleanup that? Or do you prefer a better solution and we drop this PR?

conf/layer.conf
```diff
@@ -23,6 +23,8 @@ BBFILES_DYNAMIC += " \
     \
 "
 
+LAYERDEPENDS_riscv-layer = "core openembedded-layer"
```
kraj, line 26: I would like to avoid this dep if possible. Maybe debate it with core layer to include the needed tools

lslrt: Okay, I can look at the e2fsprogs_1.47.4.bb recipes in oe-core and see if ext2simg_android can be build, configuring that with a PACKAGECONFIG.
Will eventually try to submit what I end with.

lslrt: That's seems complicated to get it into oe-core because it depends on libsparse that depends on android libbase.
I just realized that the layer dependency on openembedded-layer is kind of implicitly already there (at least if building for a machine that is supported by the layer):
indeed every kas files are including kas/base-riscv.yml which includes kas/include/oe.yml file which pull meta-oe layer from meta-openembedded.
So meta-riscv is not listing it as a layer dependency in LAYERDEPENDS but should do it because in practice none of the machine supported is building without meta-oe.
Is that really a big deal to have this layer as a dependency? What is your concern @kraj ?
I can see that some tests of check-meta-riscv are failing now that I've modified the github action to include the layer dependencies: https://github.com/lslrt/meta-riscv/actions/runs/33606680808/job/100172159145
In my last push, I've disabled the check of the dependencies, lets see if that makes the test pass.

kraj: meta-oe is included for building additional components. In this case now we are spelling it out as a hard dependency, if there are hard dependencies on meta-oe to build the BSP that would make a general case for including it, but I am not convinced thats the case right now.

PR #574, opened 2025-12-10
fixes to pass yocto check layer script

threexc:
This combination of changes should allow yocto-check-layer to pass when tested against meta-riscv. I've tested it locally like so:

```
yocto-check-layer --debug --dependency ../meta -- ../../meta-riscv 2>&1 | tee meta-riscv-check-layer.log
```

It would be good if others could also test this, however. I do know that there are unrelated issues in Python 3.14-based hosts with bitbake, so it may be best to do this test on a distro other than Fedora 43 (or in a container).

Note that if you are going to test it, you do NOT want meta-riscv specified in bblayers.conf (the `--dependency` flag tells yocto-check-layer what dependencies the layer under test expects).

### DIFF
--- a/README.md
+++ b/README.md
@@ -211,6 +211,10 @@ Instead of dding wic.gz image ```bmaptool``` (available in most Linux distributi
 $ sudo bmaptool copy <image>-<machine>.wic.gz /dev/sdX
 ```
 
+## Contributing
+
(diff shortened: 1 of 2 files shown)


kraj (approved):
LGTM

PR #577, opened 2026-01-06
meta-riscv: separate layers and build directories in repo setup

sanjay-embedded:
The current repo-based workspace places Yocto layers and the build directory at the same top-level path. This makes source navigation, searching (grep/ripgrep), and IDE indexing difficult due to build artifacts (tmp/, sstate-cache/, etc.).

This change introduces a dedicated 'layers/' directory for all Yocto layers while keeping the build directory separate:

  <workspace>/
    layers/
      poky/
      meta-riscv/
      meta-openembedded/
      openembedded-core/
    build/

Changes included:
- Update riscv-yocto.xml to fetch all layers under layers/
- Update setup.sh to support the new layout
- Update README to document the revised workspace structure

There is no functional change to the build output. Existing workflows remain compatible, and the new layout improves developer productivity and aligns with common Yocto workspace practices.

Signed-off-by: Sanjay Chitroda <sanjayembeddedse@gmail.com>

### DIFF
--- a/README.md
+++ b/README.md
@@ -37,7 +37,7 @@ Make sure to [install the `repo` command by Google](https://source.android.com/s
 ### Create workspace
 ```text
 mkdir riscv-yocto && cd riscv-yocto
-repo init -u https://github.com/riscv/meta-riscv  -b master -m tools/manifests/riscv-yocto.xml
+repo init -u https://github.com/riscv/meta-riscv -b master -m tools/manifests/riscv-yocto.xml
 repo sync
 repo start work --all
 ```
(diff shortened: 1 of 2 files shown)


kraj (changes requested):

sanjay-embedded:
Ok, understood.

May I know why do we have manifest dependency on that yoe https://github.com/riscv/meta-riscv/blob/master/tools%2Fmanifests%2Friscv-yocto.xml ?

Ideally, we can remove this dependency and clone the repo from risc-v/meta-riscv so that won't have dependency on yoe and have latest development available on master branch.

sanjay-embedded:
project name="meta-riscv" remote="yoe" path="layers/meta-riscv" revision="yoe/mut"

This should be something like.

project name="meta-riscv" remote="riscv" path="layers/meta-riscv" revision="riscv/master"

Independent own branch and remote.

If you recommend, I can review and raise pull request for the same.

Which would use meta-riscv from risc-v repo.

Issue #582, opened 2026-01-15
u-boot-spl-k1 is being pulled into qemuriscv64 builds


kraj:
This recipe was recently added, it should only be needed for k1 SOC based SBCs but its being pulled into `qemuriscv64` machine builds as well. see

https://errors.yoctoproject.org/Errors/Details/896256/

michaelopdenacker:
I'm taking care of this, thanks for notifying me!

PR #584, opened 2026-01-16
u-boot-spl-k1: make it compatible with k1 SoC only

michaelopdenacker:
By setting COMPATIBLE_MACHINE = "(k1)"
This way, other machines and SoCs don't pick it up by mistake

Addressing this issue:
https://github.com/riscv/meta-riscv/issues/582


kraj (approved):
LGTM

PR #587, opened 2026-01-18
manifest: switch remote from YoeDistro to RISC-V

### LINKED ISSUE #586: Why meta-riscv sync is dependent on Yoe Distro?
Hi @kraj ,

I want to understand why do we have manifest dependency yoe  https://github.com/riscv/meta-riscv/blob/master/tools%2Fmanifests%2Friscv-yocto.xml ?

Ideally, we can remove this dependency and clone the repo from risc-v/meta-riscv so that won't have dependency on yoe and have latest development available on master branch.

project name="meta-riscv" remote="yoe" path="layers/meta-riscv" revision="yoe/mut"

This should be something like.

project name="meta-riscv" remote="riscv" path="layers/meta-riscv" revision="riscv/master"

Independent own branch and remote.

If you recommend, I can review and raise pull request for the same which would use meta-riscv from risc-v repo.

sanjay-embedded:
- Remove the `yoe` remote along with the dependency of meta-riscv
- Add the official `riscv` remote and update meta-riscv to track the upstream master branch.

This aligns the manifest with the official RISC-V meta-riscv layer and ensures tracking of the latest upstream development.

fixes #586

### DIFF
--- a/tools/manifests/riscv-yocto.xml
+++ b/tools/manifests/riscv-yocto.xml
@@ -3,14 +3,13 @@
 <manifest>
   <remote name="openembedded" fetch="https://github.com/openembedded/" />
   <remote name="riscv" fetch="https://github.com/riscv/" />
-  <remote name="yoe" fetch="https://github.com/YoeDistro/" />
   <remote name="yocto" fetch="https://git.yoctoproject.org/" />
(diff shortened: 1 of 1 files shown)


kraj (approved):
lgtm

PR #606, opened 2026-03-12
WIP: improve K1 SoC BSPs

threexc:
In short:

1. Pull in patches from the kernel mailing list to support SD cards on the OrangePi RV2;
2. Add a new `muse-pi-pro` MACHINE for the [Muse Pi Pro](https://www.spacemit.com/community/document/info?lang=en&nodepath=hardware/eco/k1_muse_pi_pro/root_overview.md) board from SpacemiT (the main website link doesn't seem to load);
3. Consolidate all of the metadata common to `muse-pi-pro`, `orangepi-rv2-mainline`, and `bananapi-f3` into the `k1.inc` file;
4. Make some changes to `bootcommand.cfg` and `spacemit-k1.wks` to make the SD card's rootfs label deterministic;
5. Add a new `k1-image` class which should solve @michaelopdenacker 's issue with having to manually prepend `bootinfo_sd.bin` to the beginning of images before flashing.

I've tested this on my Muse Pi Pro locally and it now reliably boots a poky reference image:

```
Poky (Yocto Project Reference Distro) 5.3.99+snapshot-714ed9bc1b00a3527b822fe20677a7037923575a muse-pi-pro ttyS0

Type 'root' to login with superuser privileges (no password will be asked).

muse-pi-pro login: root

WARNING: Poky is a reference Yocto Project distribution that should be used for
testing and development purposes only. It is recommended that you create your
own distribution for production use.

root@muse-pi-pro:~#
```

and the OrangePi RV2 works too:

```
Poky (Yocto Project Reference Distro) 5.3.99+snapshot-714ed9bc1b00a3527b822fe20677a7037923575a orangepi-rv2-mainline ttyS0

Type 'root' to login with superuser privileges (no password will be asked).

orangepi-rv2-mainline login: root

WARNING: Poky is a reference Yocto Project distribution that should be used for
testing and development purposes only. It is recommended that you create your
own distribution for production use.

root@orangepi-rv2-mainline:~#
```

I may need to do a similar upstream submission for the `i2c-k1.cfg` fragment, since otherwise it only gets loaded as a module, and therefore the entire chain from the i2c bus to the MMC fails at boot.

I'm not sure if there's a better way to handle the determinism issue with the image rootfs, so suggestions are appreciated.

Thanks for reviewing!


michaelopdenacker:
Are you sure you want to merge patches that are not in mainline yet? I know we do that for some boards (like "orangepi-rv2"), but "orangepi-rv2-mainline" was intended to use mainline code only (with the exception of the SPL code for the moment).

At least as far as OrangePi RV2 is concerned, we already have a quite exhaustive version through the "orangepi-rv2" machine based on the vendor kernel...

That's not a strong opinion though. This can also help testing the patches queued for mainline inclusion.

threexc:
Pushed an update:

- use kernel 7.0-rc4 now
- pulled in v3 of Iker's upstream patch set
- refactored my Muse Pi Pro .dts in preparation for submission upstream

Tested again on the latter board and it looks OK.

Note that the series now also includes a similar SD card support patch for the BPI-F3.

ziswiler:
Yes, when I specify `mem=1G`, then it boots fine. Strangely, with `mem=2G` it still crashes, but much later. So it is clearly the memory initialisation/sizing in U-Boot which is to blame.

[opirv2-yocto.014-working-mem-1g.log](https://github.com/user-attachments/files/26110656/opirv2-yocto.014-working-mem-1g.log)

kraj (approved):
Lgtm

PR #614, opened 2026-03-24
README: fix broken VisionFive 2 links, add notes

threexc:
- There is no valid product page for the VisionFive (V1), so remove that broken link.
- Update the VisionFive 2 link.
- Add a "Discontinued" note for the freedom-u540.


kraj (approved):
LGTM. I wonder if we should just mark `v1` for removal ?

PR #621, opened 2026-03-30
k1-vendor.inc: introduce vendor include

threexc:
Switching to a common k1.inc for all of the mainline-based board recipes accidentally broke the vendor MACHINEs for the OrangePi RV2 and R2S. We are still interested in supporting these for now, so create a variant of the .inc file which matches the previous iteration but with a clear indication of its focus on the vendor-based configurations.

Thanks to @michaelopdenacker for pointing this out.

### DIFF
--- a/conf/machine/include/k1-vendor.inc
+++ b/conf/machine/include/k1-vendor.inc
@@ -10,3 +10,31 @@ require conf/machine/include/riscv/tune-riscv.inc
 require conf/machine/include/soc-family.inc
 
 SOC_FAMILY = "k1"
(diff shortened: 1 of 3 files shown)


michaelopdenacker:
It looks perfect for me, and keeps the vendor support code as clean as it can be.
It builds fine here too, and therefore we can be pretty sure the results are fine too.
I'd vote to merge this pull request.
Thanks @threexc for the fixes!

kraj (approved):
Looks good now thanks

PR #629, opened 2026-05-07
u-boot-spl-k1.bb: Use more reliable git server

EddyTheCo:
To address networking issues accessing the git server gitee.com, use Amarula clone on GitHub that is more reliable.
This addresses build issues impacting all SpacemiT k1 boards.

The issue was found and fixed on BananaPi F3.

### DIFF
--- a/recipes-bsp/u-boot/u-boot-spl-k1.bb
+++ b/recipes-bsp/u-boot/u-boot-spl-k1.bb
@@ -7,7 +7,7 @@ LIC_FILES_CHKSUM = "file://Licenses/README;md5=2ca5f2c35c8cc335f0a19756634782f1"
 
 inherit deploy
 
-SRC_URI = "git://gitee.com/bianbu-linux/uboot-2022.10;protocol=https;branch=k1-bl-v2.2.y"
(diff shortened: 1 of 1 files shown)


michaelopdenacker (commented):
I also had the same issue on Banana PI F3.
This patch fixes the build for me too.

This should be necessary for all our K1 based machines.

threexc:
How does this manifest? I'm seeing a do_patch issue with u-boot 2026.04 as of this morning for `0001-scripts-dtc-drop-yaml-in-DT-validation.patch`, but that may be unrelated.

EDIT: LGTM, but I am curious if either of you are seeing the u-boot do_patch problem I'm seeing...

EddyTheCo:
I was getting
```
fatal: unable to access 'https://gitee.com/bianbu-linux/uboot-2022.10/': Could not resolve host: gitee.com
WARNING: Failed to fetch URL git://gitee.com/bianbu-linux/uboot-2022.10;protocol=https;branch=k1-bl-v2.2.y, attempting MIRRORS if available
```
Depending on the internet provider I select, I may have a problem resolving `gitee.com`.

PR #630, opened 2026-05-07
linux-mainline-k1: misc.cfg: Remove conflicting kernel config

EddyTheCo:
Remove the configuration for CONFIG_I2C_K1 that enables it as a module because it is defined as a built-in in the k1-i2c.cfg.


michaelopdenacker:
Looks good to me. Thanks @EddyTheCo

threexc:
Yeah, LGTM.

PR #633, opened 2026-05-20
linux: remove duplicate KBUILD_DEFCONFIG definitions

michaelopdenacker:
KBUILD_DEFCONFIG ?= "defconfig" is sufficient in
recipes-kernel/linux/linux-mainline-common.inc,
making identical settings redundant.

Similarly, this setting is not necessary in
linux-mainline-k1.bb which includes
linux-mainline-common.inc.


threexc:
LGTM. Thanks!

kraj (approved):

PR #636, opened 2026-05-21
Add SpacemiT K3 Vendor BSP

threexc:
## Overview

This is an effort on behalf of SpacemiT to add support for their K3 platforms to meta-riscv. It breaks down the changes made in the [spacemit-yocto](https://github.com/yingjie-liu-spacemit/spacemit-yocto) fork of meta-riscv into more discrete and digestible pieces focused on a minimal, headless image, while omitting the parts which aren't strictly required, notably:

- Other boards (will be added on a case-by-case basis, as people can test them)
- Custom toolchain usage for firmware blobs
- Changes to the `mesa` recipes and related graphics support
- Support for the SpacemiT Titan flashing process

These features will be investigated after basic support is tested and merged.

I've opened this now so that others can review and propose additional BSP support if they have other K3 boards to test with. I've followed the same format we did for the K1 platforms and created a `k3-vendor.inc` file with the common logic included so that adding those other BSPs as we go will require minimal effort.

## Status

**As of EOD on May 29th, 2026**, I have confirmed that this works by booting three ways:

1. kernel + dtb + Bianbu rootfs (from UFS):

```
Bianbu 4.0rc2 tgamblin-spacemitk3picoitx ttyS0
tgamblin-spacemitk3picoitx login: tgamblin
Password:
Welcome to Bianbu 4.0rc2 (GNU/Linux 6.18.3-k3 riscv64)

 * Documentation:  https://bianbu.spacemit.com
 * Support:        https://ticket.spacemit.com

tgamblin@tgamblin-spacemitk3picoitx:~$
```

2. kernel + dtb + initramfs via `tftpboot`:

```
~ # uname -a
Linux (none) 6.18.3-k3 #1 SMP PREEMPT_DYNAMIC Thu Apr 30 08:02:53 UTC 2026 riscv64 GNU/Linux
~ # cat /etc/issue
OpenEmbedded nodistro.0 \n \l
~ # dmesg | grep -i riscv
[    0.000000] Linux version 6.18.3-k3 (oe-user@oe-host) (riscv64-oe-linux-gcc (GCC) 15.2.0, GNU ld (GNU Binutils) 2.46) #1 SMP PREEMPT_DYNAMIC Thu Apr 30 08:02:53 UTC 2026
[    0.000000] riscv: base ISA extensions acdfhimv
[    0.000000] riscv: ELF capabilities acdfimv
[    0.000000] riscv-intc: 64 local interrupts mapped using AIA
[...]

PR #637, opened 2026-05-22
u-boot: add u-boot-spl-spacemit.inc, use in u-boot-spl-k1

threexc:
Rework the u-boot-spl-k1 recipe to store the generic functions and repo metadata in a new 'u-boot-spl-spacemit.inc' file, then use the require directive to include that logic in the much-simplified u-boot-spl-k1 recipe. This allows the common elements to be re-used for new BSPs, such as those for the SpacemiT's K3 platforms, which follow the same basic SPL usage pattern.

---

This is a preparatory change for #636 - the early implementation I'm working off of there tries to build the SPL parts in the same recipe as the main U-Boot output, which is at odds with how we currently do things for the K1, and which also requires a non-ideal workaround patch.

If this should be done differently, let me know. I had considered changing `u-boot-spl-k1.bb` to `u-boot-spl.bb` and doing machine-specific variables and functions, but I thought this was clearer and easier to maintain as we go - especially when we add similar BSPs for other K3 board variants (and the future K5 and K7 platforms mentioned [on Reddit](https://www.reddit.com/r/spacemit_riscv/s/Oz2Jx1Sf5U)), especially if they have slight differences in which files they need to deploy.

I've tested this for regressions on my local infra with my BananaPi F3, which uses Labgrid with an SD Mux to let me image and boot the board remotely. I've started uploading these test results to a public mirror - you can see the Labgrid pytest report for the run [here](https://testresults.ecocode.ca/yocto/meta-riscv/bananapi-f3/20260522/1171/report.html?sort=result)

### DIFF
--- a/recipes-bsp/u-boot/u-boot-spl-k1.bb
+++ b/recipes-bsp/u-boot/u-boot-spl-k1.bb
@@ -1,41 +1,12 @@
-HOMEPAGE = "https://gitee.com/bianbu-linux/uboot-2022.10"
-DESCRIPTION = "U-Boot SPL for SpacemiT K1 SoCs, derived from U-Boot"
-SECTION = "bootloaders"
-DEPENDS += "flex-native bison-native python3-setuptools-native"
-LICENSE = "GPL-2.0-or-later"
-LIC_FILES_CHKSUM = "file://Licenses/README;md5=2ca5f2c35c8cc335f0a19756634782f1"
+require u-boot-spl-spacemit.inc
 
-inherit deploy
-
-SRC_URI = "git://github.com/amarula/uboot-2022.10;protocol=https;branch=k1-bl-v2.2.y"
-SRCREV = "c6f2746cb7993a6fb6c9f51b2bff318921e13f98"
-
-UBOOT_SPL_MACHINE ??= "k1_defconfig"
-
-EXTRA_OEMAKE = "\
-    CROSS_COMPILE=${TARGET_PREFIX} \
-    ARCH=riscv \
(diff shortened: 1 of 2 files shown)


michaelopdenacker:
Looks like a good idea!

recipes-bsp/u-boot/u-boot-spl-k1.bb
```diff
@@ -1,41 +1,12 @@
-HOMEPAGE = "https://gitee.com/bianbu-linux/uboot-2022.10"
```
alistair23, line 1: Not sure about the name here, you plan to use this with the k3, but the file is called  `u-boot-spl-k1.bb`

If it's just for the K1 and K3 I'm not sure this refactor is worthwhile. Maybe one day we can use upstream u-boot for one of those and then we don't really need the extra include

threexc: The `u-boot-spl-k1.bb` file is the original name, I just put most of its contents into the `.inc` file so that I could then re-use those for the `k3` BSP there. I'm OK with not doing this and having two files, but they will be very similar.

kraj: Common file looks a good idea. K3 recipe Will come as part of k3 bsp PR I guess

threexc: Correct, that branch currently sits on top of this one with a corresponding k3 recipe. I will rebase them both shortly.

kraj (approved):

PR #639, opened 2026-05-24
machine: add bananapi-cm6-io

michaelopdenacker:
This adds support for the BananaPI CM6 Compute Module in its IO board.

At this stage, Linux kernel support is provided through separate DT patches submitted by Junhui Liu <junhui.liu@pigmoral.tech> to the mainline Linux kernel.


michaelopdenacker:
I supplied the DT patches separately for this new board, because they are separate from Iker Pedrosa's patches. Both sets of patches will be easier to maintain if they are separate.

Ultimately of course, these patches will go away.

docs/bananapi-cm6-io.md
```diff
@@ -0,0 +1,97 @@
...
+* openembedded-core
+  * meta
+* meta-riscv
+
+See [the Yocto Project](https://docs.yoctoproject.org/brief-yoctoprojectqs/index.html) manual for details.
+
+Set these variables in a configuration file:
+
+* `MACHINE = "bananapi-cm6-io"`
+* `DISTRO = "poky-altcfg"`
```
alistair23, line 24: You don't mention poky as a requirement

docs/bananapi-cm6-io.md
```diff
@@ -0,0 +1,97 @@
...
+* `EXTRA_IMAGE_FEATURES = "allow-empty-password empty-root-password allow-root-login post-install-logging"`
+
+Build your image:
+
+```
+$ bitbake core-image-full-cmdline
```
alistair23, line 30: I think this should just be dropped from the board specific doc

michaelopdenacker:
Hi @alistair23 
Thanks for the review!
Oops, it seems that GitHub swallowed my answers to your comments...
Actually, the documentation bits you point at are also part of the documentation for BananaPI F3, Orange PI RV2 and R2S and maybe other boards.
What about proposing a separate change to address all such documentation bits at once?

kraj (approved):
lgtm

kraj:
> Hi @alistair23 Thanks for the review! Oops, it seems that GitHub swallowed my answers to your comments... Actually, the documentation bits you point at are also part of the documentation for BananaPI F3, Orange PI RV2 and R2S and maybe other boards. What about proposing a separate change to address all such documentation bits at once?

If you plan to do it in a followup that will be fine I think. @alistair23 what do you think?

threexc:
LGTM. Thanks for submitting this. Unfortunately I can't test it myself, but it looks good!

As an aside: I notice that here (and in your dev day talks this morning) you mention a manual process to write bootinfo. Can you not use the `k1-image` class and the `spacemit-k1.wks` (or a new one, if needed) to avoid the manual steps?

michaelopdenacker:
> LGTM. Thanks for submitting this. Unfortunately I can't test it myself, but it looks good!
> 
> As an aside: I notice that here (and in your dev day talks this morning) you mention a manual process to write bootinfo. Can you not use the `k1-image` class and the `spacemit-k1.wks` (or a new one, if needed) to avoid the manual steps?

Hi Trevor
Thanks for the suggestion, I'll have a look!

PR #640, opened 2026-05-24
linux-mainline-k1: remove NFS usage

michaelopdenacker:
NFS no longer necessary thanks to the MMC support patches. Also remove NFS from usage guidelines.


alistair23 (approved):

docs/bananapi-f3.md
```diff
@@ -75,45 +75,6 @@ Then, you also need to flash the `bootinfo_sd.bin` file at the very beginning of
 $ sudo dd if=build/tmp/deploy/images/bananapi-f3/bootinfo_sd.bin of=/dev/sdx
 ```
 
-Set up NFS
```
kraj, line 78: I see that its not necessary anymore but why not keep this information available optionally, so if someone wants to use nfsroot then they have a place to start ?

michaelopdenacker: Then we would have to explain how to add back NFS settings in the kernel command line. IMHO, this becomes complicated. Overly complicated documentation could scare beginners.

kraj: Ok I guess. Start simple is better

kraj (approved):

PR #641, opened 2026-05-24
CONTRIBUTORS: update

michaelopdenacker:
The list of people was generated with:
git log | grep Signed | sort | uniq


michaelopdenacker:
Another possibility is to remove this file, to avoid having to maintain it :D
If we decide to do so, we also have to stop mentioning this file in the "LICENSE" file.

threexc:
> Another possibility is to remove this file, to avoid having to maintain it :D If we decide to do so, we also have to stop mentioning this file in the "LICENSE" file.

I'd be OK with removing it. Between the GitHub web UI and `git shortlog -sne`, there are three ways already to track who's contributed. If we were going to keep it, maybe we should add some sort of CI workflow to automatically update a list?

kraj:
Yeah let's remove it. Stats in github can provide this information. Its painful to keep this file updated

kraj (changes requested):
Lets delete this file

kraj:
> Lets delete this file

@threexc please revise this PR as well

threexc:
@kraj did you mean @michaelopdenacker ?

michaelopdenacker:
Done. I removed the file.
This should be ready to merge according to previous discussions.

kraj (approved):
LGTM

PR #645, opened 2026-06-03
opensbi: Fixup build failures with C23 enabled compilers

alistair23:


kraj (approved):
Lgtm

OldManYellsAtCloud:
I just wanted to check if there is no requirement for a bracket after the word `Backport` in the patch file, but while doing so
- the kernel failed to patch
- once I fixed that, it failed to compile
- once I fixed that, u-boot failed to compile

I think star64 is dead :(

threexc:
> I just wanted to check if there is no requirement for a bracket after the word `Backport` in the patch file, but while doing so
> 
>     * the kernel failed to patch
> 
>     * once I fixed that, it failed to compile
> 
>     * once I fixed that, u-boot failed to compile
> 
> 
> I think star64 is dead :(

If it is, we should add it to the deprecated list. Let's try to confirm there's nothing we can do first though. I'm travelling but I'll try taking a look over the next few days.

PR #647, opened 2026-06-04
Initial support for dc-roma-fml13v01 with WiFI

alistair23:
This is initial support for the [DC-ROMA RISC-V Mainboard](https://deepcomputing.io/product/dc-roma-risc-v-mainboard/) with WiFi and Ethernet enabled.


threexc:
Looks OK to me at first glance. Do you have an idea of the overlap between this and the DC Roma II? Given that there's also a DC Roma III out there, I wonder if it'd make sense to create a `.inc` for some pieces.

alistair23:
> Looks OK to me at first glance. Do you have an idea of the overlap between this and the DC Roma II? Given that there's also a DC Roma III out there, I wonder if it'd make sense to create a `.inc` for some pieces.

I think they are completely different. They are different SoCs  JH7110 (Roma 1) vs ESWIN EIC7X (Roma 2) vs K3 (Roma 3) from different manufactures. The form factor seems to be the main similarity.

I don't think we can share a lot. We are better off sharing between similar boards (for example the Roma 3 could share with other K3 powered boards).

threexc:
> > Looks OK to me at first glance. Do you have an idea of the overlap between this and the DC Roma II? Given that there's also a DC Roma III out there, I wonder if it'd make sense to create a `.inc` for some pieces.
> 
> I think they are completely different. They are different SoCs JH7110 (Roma 1) vs ESWIN EIC7X (Roma 2) vs K3 (Roma 3) from different manufactures. The form factor seems to be the main similarity.
> 
> I don't think we can share a lot. We are better off sharing between similar boards (for example the Roma 3 could share with other K3 powered boards).

Thanks. I figured that'd be the case for the Roma 3, but good to confirm for the others too.

recipes-kernel/linux/linux-mainline/ethernet.cfg
```diff
@@ -0,0 +1,2 @@
+CONFIG_USB_RTL8150=m
+CONFIG_USB_RTL8152=m
```
kraj, line 2: I wonder if it's better to call it something specific for rtl or somesuch instead of `ethernet.cfg`

recipes-kernel/linux/linux-mainline/wifi.cfg
```diff
@@ -0,0 +1,6 @@
+CONFIG_CFG80211=m
+CONFIG_MAC80211=m
+CONFIG_WLAN=y
+CONFIG_WLAN_VENDOR_INTEL=y
```
kraj, line 4: same comment here as well.

recipes-kernel/linux/linux-mainline/ethernet.cfg
alistair23: It's used to support the [Framework Ethernet Expansion Card](https://frame.work/au/en/products/ethernet-expansion-card), so I do think `ethernet.cfg` is the correct name

alistair23: How about `framework-ethernet.cfg`?

PR #650, opened 2026-06-12
machine: add visionfive2-mainline

mkaehlcke:
This adds support for booting the upstream Linux kernel and U-Boot on the StarFive VisionFive 2 (and Milk-V Mars)

At this stage the board boots to the Linux console, support for components that require non-upstream drivers is currently not included.

This adds the machine includes visionfive2-common.inc and jh7110-vendor.inc to avoid mostly redundant configurations for the visionfive2 and the visionfive2-mainline boards.

The new kernel recipe linux-mainline-jh7110.bb has been added for boards based on the StarFive JH-7110 SoC.

Testing involved building core-images for visionfive2-mainline and visionfive2 and booting them on a Milk-V Mars board. It has been confirmed that the variables that specify the Linux kernel and U-Boot configuration for the existing visionfive2 board were not altered by the consolidation of the visionfive2/jh7110 configuration.

Signed-off-by: Matthias Kaehlcke <matthias@kaehlcke.net>

### DIFF
--- a/conf/machine/include/jh7110-vendor.inc
+++ b/conf/machine/include/jh7110-vendor.inc
@@ -1,0 +1,34 @@
+#@TYPE: SOC
+#@NAME: StarFive JH7110
+#@SOC: StarFive JH7110
+
+#============================================
+# Common Preferred Providers
+PREFERRED_PROVIDER_virtual/kernel ?= "linux-starfive-dev"
+PREFERRED_PROVIDER_virtual/bootloader ?= "u-boot-starfive"
+PREFERRED_PROVIDER_virtual/libgl ?= "mesa-pvr"
+PREFERRED_PROVIDER_virtual/libglx ?= "mesa-pvr"
+PREFERRED_PROVIDER_virtual/mesa ?= "mesa-pvr"
+PREFERRED_PROVIDER_virtual/libgbm ?= "mesa-pvr"
+PREFERRED_PROVIDER_virtual/egl ?= "mesa-pvr"
+PREFERRED_PROVIDER_virtual/libgles3 ?= "visionfive2-pvr-graphics"
+PREFERRED_PROVIDER_virtual/libgles1 ?= "visionfive2-pvr-graphics"
+PREFERRED_PROVIDER_virtual/libgles2 ?= "visionfive2-pvr-graphics"
+PREFERRED_PROVIDER_virtual/libomxil ?= "libsf-omxil"
+#============================================
+
+#============================================
+# Common Machine Configuration
+MACHINE_ESSENTIAL_EXTRA_RRECOMMENDS += " \
+    linux-firmware-visionfive2-imggpu \
+    vdec-module \
+    venc-module \
+    jpu-module \
+    visionfive2-pvr-graphics-tools \
+"
+#============================================
+
+#============================================
+# Common Kernel Configuration
+KERNEL_MODULE_AUTOLOAD += "pvrsrvkm"
+#============================================
--- a/conf/machine/visionfive2-mainline.conf
+++ b/conf/machine/visionfive2-mainline.conf
@@ -1,0 +1,8 @@
+#@TYPE: Machine
+#@NAME: visionfive2-mainline
+#@SOC: StarFive JH7110
+#@DESCRIPTION: Machine configuration for the VisionFive 2 board with mainline Linux and U-Boot
+
+require include/visionfive2-common.inc
+
+PREFERRED_PROVIDER_virtual/kernel ?= "linux-mainline"
--- a/README.md
+++ b/README.md
@@ -53,6 +53,7 @@ This layer depends on:
 | star64                   | [PINE64 STAR64](https://pine64.org/devices/star64/)                                                          | 5.15 Kernel [fork](https://github.com/fishwaldo/Star64_linux)                    |
 | visionfive               | StarFive VisionFive                                                                                          | No product page found                                                            |
 | visionfive2              | [StarFive VisionFive 2](https://www.starfivetech.com/en/index.php?s=hardware&c=show&id=14)                   |                                                                                  |
+| visionfive2-mainline     | [StarFive VisionFive 2](https://www.starfivetech.com/en/index.php?s=hardware&c=show&id=14)                   | Mainline Kernel and U-Boot                                                       |
 
 Note that this layer also provides improvements and features for the
 upstream qemuriscv32 and qemuriscv64 machines.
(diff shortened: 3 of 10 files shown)


alistair23:
Thanks for the patch!

After this PR `linux-starfive-dev` is only used for the visionfive and beaglev, both of which are deprecated. So I think it's worth removing the `visionfive2` and `jh7110` from `linux-starfive-dev`. Then `linux-starfive-dev` is ready to just be deleted once we remove visionfive and beaglev.

So basically let's transition to only using the mainline kernel for the visionfive2

conf/machine/include/jh7110-vendor.inc
```diff
@@ -0,0 +1,35 @@
+#@TYPE: SOC
+#@NAME: StarFive JH7110
+#@SOC: StarFive JH7110
+#@DESCRIPTION: Common config for boards based on the StarFive JH7110 - vendor repo version
```
alistair23: Just remove this

MaxKrummenacher:
> So basically let's transition to only using the mainline kernel for the visionfive2

If I got @mkaehlcke 's comments right the mainline kernel currently lacks a lot of functionality, so removing the option to use the vendor kernel removes support for a lot of use cases.

threexc:
> > So basically let's transition to only using the mainline kernel for the visionfive2
> 
> If I got Matthias Kaehlcke (@mkaehlcke) 's comments right the mainline kernel currently lacks a lot of functionality, so removing the option to use the vendor kernel removes support for a lot of use cases.

Yes, we generally want to keep vendor BSPs until we're sure that upstream has maximal support merged, at which point we can deprecate the vendor versions. There should also come a point at which we can put most or all mainline kernel variants into the same recipe.

As a side note (not essential for this PR specifically), it would be interesting to see a comparison of features for the visionfive2 and other boards...

conf/machine/include/jh7110-vendor.inc
mkaehlcke: done in the latest push

mkaehlcke:
> > > So basically let's transition to only using the mainline kernel for the visionfive2
> > 
> > 
> > If I got Matthias Kaehlcke (Matthias Kaehlcke (@mkaehlcke)) 's comments right the mainline kernel currently lacks a lot of functionality, so removing the option to use the vendor kernel removes support for a lot of use cases.
> 
> Yes, we generally want to keep vendor BSPs until we're sure that upstream has maximal support merged, at which point we can deprecate the vendor versions. There should also come a point at which we can put most or all mainline kernel variants into the same recipe.
> 
> As a side note (not essential for this PR specifically), it would be interesting to see a comparison of features for the visionfive2 and other boards...

Thanks for the reviews!

I did not remove linux-starfive-dev for now, as the upstream kernel still lacks functionality, among other things in the graphics department.

recipes-kernel/linux/linux-mainline-jh7110.bb
```diff
@@ -0,0 +1,13 @@
...
+
+require recipes-kernel/linux/linux-mainline-common.inc
+
+DEPENDS += "u-boot-tools-native"
+
+BRANCH = "master"
+SRC_URI = "git://git.kernel.org/pub/scm/linux/kernel/git/torvalds/linux.git;protocol=https;branch=${BRANCH} \
+          "
+SRCREV = "028ef9c96e96197026887c0f092424679298aae8"
+LINUX_VERSION = "7.0"
```
alistair23: Can't you just use `linux-mainline_7.0`?

conf/machine/visionfive2-mainline.conf
```diff
@@ -0,0 +1,8 @@
+#@TYPE: Machine
+#@NAME: visionfive2-mainline
+#@SOC: StarFive JH7110
+#@DESCRIPTION: Machine configuration for the VisionFive 2 board with mainline Linux and U-Boot
```
alistair23, line 4: I think we should just support `PREFERRED_PROVIDER` instead of a new machine.

Assuming you don't want to support a mainline only kernel and drop the vendor support

alistair23 (changes requested):

recipes-kernel/linux/linux-mainline-jh7110.bb
threexc: This is a growing problem in the layer, IMO. I'm intending on submitting a PR to combine the mainline kernel recipes later this week, provided that no one else beats me to it. For now I'd be OK with this as-is, since I'll refactor it (it'll just mean a bit of churn). However, if it could be added to the existing mainline recipe, then that'd make things easier.

mkaehlcke: > Can't you just use `linux-mainline_7.0`?

In general that would be an option

If in the future someone wanted to add jh7110/visionfive2 specific patches I suppose they could do this through overrides, or would it be preferred to have a dedicated kernel recipe in that case?

conf/machine/visionfive2-mainline.conf
mkaehlcke: Would it be somehow possible to only apply the settings in conf/machine/include/jh7110-vendor.inc if PREFERRED_PROVIDER is set to linux-starfive-dev?

recipes-kernel/linux/linux-mainline-jh7110.bb
alistair23: > If in the future someone wanted to add jh7110/visionfive2 specific patches I suppose they could do this through overrides, or would it be preferred to have a dedicated kernel recipe in that case?

Yeah, you can just use a `SRC_URI:append:xxx` to add a few board specific patches to the kernel. Consolidating on a single "mainline" kernel just makes it easier to update. Otherwise this 7.0 release is going to get old quickly as it isn't getting any patch updates

mkaehlcke: > Can't you just use `linux-mainline_7.0`?

done :)

conf/machine/visionfive2-mainline.conf
kraj: well, selecting which recipe provides kernel is not going to give you a knob to check something else based on this information, machines add an override which is handy to use. However, it is subject to kind of changes expected for it to work. If its just kernel toggling then PREFERRED_PROVIDER is a good approach. More than that then it get trickier to keep supporting two kernels on same machine type. Maybe another DISTRO variable to denote this information has to be added.

mkaehlcke: Are you thinking about a DISTRO variable that indicates that a vendor or mainline kernel is used? And the machine configuration would chose the correct settings based on the value of that variable?

PR #659, opened 2026-07-30
linux-mainline-k1: update kernel to 7.2-rc5

fcastagnotto:
# Description

Update linux-mainline-k1 to Linux 7.2-rc5.

Patches already included in the updated kernel are removed.

The RISC-V vector unaligned access probe crashes during boot on the
SpacemiT K1 while executing __riscv_copy_vec_words_unaligned().
Conservatively mark vector unaligned accesses as slow to prevent the
runtime probe from running, while retaining RISC-V Vector support.

## Checklist

- [x] I have read and understood the [Contributing Changes to a Component](https://docs.yoctoproject.org/dev/contributor-guide/submit-changes.html) and [Recipe Style Guide](https://docs.yoctoproject.org/dev/contributor-guide/recipe-style-guide.html#patch-upstream-status) sections in the Yocto Project documentation
- [x] I have tested this change (provide brief details below)
- [x] Documentation has been added/updated where necessary in the README and `docs/`
- [ ] If the change is to a BSP in
  [DEPRECATED.md](https://github.com/riscv/meta-riscv/blob/master/DEPRECATED.md),
  then it is maintenance-only, i.e. it does not add support for a
  previously-removed BSP (or significant features to one slated for removal)
- [ ] (For new/modified BSPs) I have added relevant information in the README [table](https://github.com/riscv/meta-riscv/tree/master?tab=readme-ov-file#available-machines)
- [ ] (For new/modified BSPs) The bitbake-setup templates in `bitbake-registry/`
  have been updated, if necessary
- [ ] (For new/modified BSPs) A `kas` file has been provided in `kas/`
- [x] I have added my `Signed-off-by` and any other tags to each commit

# How has this been tested?

I tested the change in two ways:

- I performed a complete Yocto build for the Orange Pi RV2, deployed the resulting image to the board, and verified that it booted successfully.
- I manually configured and built Linux 7.2-rc5 using the configuration fragment, installed the resulting kernel in the SD card's boot BLOB, and successfully booted it on an Orange Pi RV2.

In both cases, I performed multiple reboots from the shell and verified that the board booted successfully each time.


recipes-kernel/linux/linux-mainline-k1.bb
```diff
@@ -8,24 +8,19 @@ BRANCH = "master"
...
-           file://0003-mmc-sdhci-of-k1-add-regulator-and-pinctrl-voltage-sw.patch \
-           file://0004-mmc-sdhci-of-k1-add-comprehensive-SDR-tuning-support.patch \
-           file://0005-riscv-dts-spacemit-k1-orangepi-rv2-add-PMIC-and-powe.patch \
-           file://0006-riscv-dts-spacemit-k1-orangepi-rv2-add-SD-card-suppo.patch \
-           file://0007-riscv-dts-spacemit-k1-bananapi-f3-add-SD-card-suppor.patch \
-           file://0008-riscv-dts-spacemit-k1-musepi-pro-add-SD-card-support.patch \
-           file://0009-riscv-dts-spacemit-k1-add-SD-card-controller-and-pin.patch \
           "
 
+SRC_URI:append:orangepi-rv2-mainline = " \
```
michaelopdenacker: This change doesn't look specific to Orange Pi RV2.
I believe you should keep it with the other changes, next to k1-i2c.cfg and misc.cfg

threexc: Yes, it should definitely be generic and not use the MACHINE override. I think it would also be good if that patch and recipe change were included as a separate commit from the uprev itself.

threexc:
Other than my response to @michaelopdenacker's thread about the kernel config, this mostly LGTM. I tested it in my CI system and it looked OK.

It looks like you removed the PR checklist instead of filling it out?

recipes-kernel/linux/linux-mainline-k1.bb
fcastagnotto: I can split the uprev and the configuration workaround into two commits. However, please note that the intermediate uprev commit builds but does not boot on the K1 without the configuration fragment. Is that the preferred structure?

fcastagnotto:
> Other than my response to Michael Opdenacker (@michaelopdenacker)'s thread about the kernel config, this mostly LGTM. I tested it in my CI system and it looked OK.
> 
> It looks like you removed the PR checklist instead of filling it out?

Sorry @threexc I didn’t notice the PR template when I opened the pull request. I’ll restore and complete it soon. It wasn’t intentional.

michaelopdenacker:
Tested successfully here. The generated image boots fine on OrangePi RV2.
Just waiting for your updates.
Thanks a lot @fcastagnotto!

recipes-kernel/linux/linux-mainline-k1.bb
michaelopdenacker: Good question: can we (knowingly) introduce a commit that breaks booting on an existing board?

threexc: Right, the intermediate version won't boot. I should've clarified on why I thought they should be split - the two commits could normally be squashed on merge anyway, but splitting them could make it easier to address other review feedback for the config files. For this PR it's probably OK to leave it.

fcastagnotto:
Thanks for the feedback. I’ve restored and completed the PR template and updated the recipe accordingly, moving the configuration fragment alongside the other generic fragments.

michaelopdenacker:
Looking good to me now. Thanks @fcastagnotto

kraj (approved):
LGTM

PR #660, opened 2026-07-30
milkv-duo-fsbl: fix build with newer binutils

MaxKrummenacher:
With ld 2.47 building fails with the following error:

|   LD      build/cv180x/bl2/bl2.elf
| riscv64-tdx-linux-musl-ld -o build/cv180x/bl2/bl2.elf -Os --gc-sections   -Map=./build/cv180x/bl2/bl2.map --script ./build/cv180x/bl2/bl2.ld ./build/cv180x/bl2/build_message.o plat/cv180x/bl2_objs/cv1800b_milkv_duo_sd/bl2/*.o ./build/cv180x/bl2/bl2_main.o ./build/cv180x/bl2/bl2_entrypoint.o
| .../riscv64-tdx-linux-musl-ld: -O requires a numerical argument or '-' or 'fast' or 'default'make: *** [Makefile:206: build/cv180x/bl2/bl2.elf] Error 1

Older ld seem to have silently ignored the undefined cmdline parameter -Os.

# How has this been tested?

A build of a simple image boots rom->fsbl->opensbi->u-boot. I.e. works as before on a milkV Duo.

### DIFF
--- a/recipes-bsp/milkv-duo-fsbl/milkv-duo-fsbl.bb
+++ b/recipes-bsp/milkv-duo-fsbl/milkv-duo-fsbl.bb
@@ -4,7 +4,10 @@ LIC_FILES_CHKSUM = "file://${COMMON_LICENSE_DIR}/Proprietary;md5=0557f9d92cf58f2
 
 inherit nopackages deploy
 
-SRC_URI = "git://github.com/milkv-duo/milkv-duo-buildroot-libraries;protocol=https;branch=main"
+SRC_URI = " \
+    git://github.com/milkv-duo/milkv-duo-buildroot-libraries;protocol=https;branch=main \
+    file://0001-milkv-duo-fsbl-fix-build-with-newer-binutils.patch;patchdir=${UNPACKDIR}/${BP} \
+"
 SRCREV = "f359994bd497f942bb67734280d81f6640c7c168"
 
 COMPATIBLE_MACHINE = "milkv-duo"
--- a/recipes-bsp/milkv-duo-fsbl/milkv-duo-fsbl/0001-milkv-duo-fsbl-fix-build-with-newer-binutils.patch
+++ b/recipes-bsp/milkv-duo-fsbl/milkv-duo-fsbl/0001-milkv-duo-fsbl-fix-build-with-newer-binutils.patch
@@ -1,0 +1,34 @@
+From 26f3cbb8c0dfbf4f0fe759c142e0633394888181 Mon Sep 17 00:00:00 2001
+From: Max Krummenacher <max.oss.09@gmail.com>
+Date: Thu, 30 Jul 2026 13:39:11 +0000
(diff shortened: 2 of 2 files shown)


kraj (commented):
Did -Os work in the past ? If so how does it compare with O1 ? Does it boot fine

MaxKrummenacher:
> Did -Os work in the past ? If so how does it compare with O1 ? Does it boot fine

AFAIK "-Os" was never a valid setting. However older ld did ignore it, linked anyway and the resulting milkv-duo-fsbl worked on the target.
ld 2.47 prints now an error and refuses to link.

https://sourceware.org/binutils/docs/ld.html#index-_002dO-level

kraj:
> > Did -Os work in the past ? If so how does it compare with O1 ? Does it boot fine
> 
> AFAIK "-Os" was never a valid setting. However older ld did ignore it, linked anyway and the resulting milkv-duo-fsbl worked on the target. ld 2.47 prints now an error and refuses to link.
> 

Understood that. Question is that this patch is changing it to -O1 and is this the default behavior that was implicitly happening with old linker?

I would rather not use any optimizations if that is how it worked before 

> https://sourceware.org/binutils/docs/ld.html#index-_002dO-level

MaxKrummenacher:
With 2.47 we get the following binary sizes:
```
-O0
-rwxr-xr-x. 1 mk users 47128 31. Jul 11:27 bl2.bin
-rwxr-xr-x. 1 mk users 3202088 31. Jul 11:27 bl2.elf

-O1 
-rwxr-xr-x. 1 mk users 47144 31. Jul 11:16 bl2.bin
-rwxr-xr-x. 1 mk users 2214480 31. Jul 11:16 bl2.elf

No "-Ox" option
-rwxr-xr-x. 1 mk users 47144 31. Jul 11:20 bl2.bin
-rwxr-xr-x. 1 mk users 2214480 31. Jul 11:20 bl2.elf
```

So omitting the option defaults to '-O1'.

Reverting oe-core's binutils update 2.46->2.47, i.e. using 2.46:
```
-O0
-rwxr-xr-x. 1 mk users 47144 31. Jul 12:28 bl2.bin
-rwxr-xr-x. 1 mk users 2214480 31. Jul 12:28 bl2.elf

-O1
-rwxr-xr-x. 1 mk users 47144 31. Jul 12:27 bl2.bin
-rwxr-xr-x. 1 mk users 2214480 31. Jul 12:27 bl2.elf

No "-Ox" option
-rwxr-xr-x. 1 mk users 47144 31. Jul 11:35 bl2.bin
-rwxr-xr-x. 1 mk users 2214480 31. Jul 11:35 bl2.elf
```
So in 2.46 the option had no effect on the resulting artifact sizes and it looks like it did the same as the 2.47 default or '-O1' does.

I will change the patch to delete the offending '-Os' rather than changing to a valid setting.

kraj (approved):
LGTM

